PR #8, opened 2018-08-16
[login-changes] Begin to support multiple Salesforce login endpoints

wlonk:
red panda


wlonk:
Note that commit 0bfbccc says the following:

You will need to do some manual database munging here, because allauth doesn't handle a settings-change gracefully.

In `python manage.py dbshell`:

    DROP TABLE account_emailaddress CASCADE;
    DROP TABLE account_emailconfirmation CASCADE;
    DELETE FROM django_migrations WHERE app = 'account';

Then re-create the emails tables:

    python manage.py migrate

This may break your existing users, too, in which case you'll want to destroy and recreate them. In `python manage.py shell`:

    from django.contrib.auth import get_user_model
    User = get_user_model()
    User.objects.all().delete()

PR #15, opened 2018-08-23
[fixgunicorn] Include gunicorn and update procfile for heroku deployment

cdcarter:


app.json
```diff
@@ -0,0 +1,50 @@
...
+  ],
+  "buildpacks": [
+      {"url": "heroku/node"},
+      {"url": "heroku/python"}
+  ],
+  "environments": {
+    "test": {
+      "scripts": {
+        "test-setup": "pip install --upgrade -r requirements/test.txt",
+        "test": "python manage.py test"
```
wlonk: Do we want to add JS tests in here? @jgerigmeyer ?

jgerigmeyer: I'm not really sure what the goal is -- are we using Heroku CI instead of CircleCI? If so, yes we can add JS tests here.

Related: Is this the command we're using to run pytests? I didn't think pytests were setup yet -- at least they're not running on CircleCI at the moment.

wlonk: No, that's a good point; this will run, but trigger no tests. Meanwhile, `pytest` will run two things that aren't tests but that it thinks are tests, so I'm gonna patch that.

app.json
```diff
@@ -0,0 +1,50 @@
...
+    "test": {
+      "scripts": {
+        "test-setup": "pip install --upgrade -r requirements/test.txt",
+        "test": "pytest"
+      },
+      "env": {
+        "DJANGO_SETTINGS_MODULE": "config.settings.test",
+        "DATABASE_URL": "sqlite:///test.db"
+      }
+    }
```
jgerigmeyer: I'm still not sure what the goal is here -- do we want to use Heroku CI? Instead of CircleCI?

wlonk: So, I don't know the answer to that, either. I think that's up to @cdcarter.

cdcarter: we can keep using CircleCI for development, but we definitely want default settings here if someone else were to want to deploy the app and use HerokuCI. they don't have to be comprehenzsive but they shouldn't be broke

jgerigmeyer: The docs make it sound like the Node buildpack automatically runs `yarn test` in Heroku CI... so I'm not sure anything else is needed here? Not 100% sure without enabling Heroku CI on the pipeline to test it.

PR #17, opened 2018-08-24
[select-products] Select products from list

jgerigmeyer:
https://trello.com/c/rN9SRmI6

axolotl

### DIFF
--- a/src/js/components/products.js
+++ b/src/js/components/products.js
@@ -1,0 +1,57 @@
+// @flow
+
+import * as React from 'react';
+import Avatar from '@salesforce/design-system-react/components/avatar';
+import Card from '@salesforce/design-system-react/components/card';
+import { Link } from 'react-router-dom';
+
+type ProductType = {
+  +id: number,
+  +name: string,
+  +version: number,
+  +body: string,
+};
+
+const Product = ({ item }: { item: ProductType }) => (
+  <Link
+    to={`/products/${item.id}`}
+    className="slds-text-link_reset slds-p-around_medium"
+  >
+    <Card
+      heading={item.name}
+      icon={<Avatar variant="entity" label={item.name} />}
+    >
+      <div className="slds-card__body_inner">
+        <div className="slds-text-title">Version {item.version}</div>
+        <p>{item.body}</p>
+      </div>
+    </Card>
+  </Link>
+);
+
+const Products = () => {
+  const fakeData = [
+    {
+      id: 1,
+      name: 'Product 1',
+      version: 3.13,
+      body: 'This is a description of the product.',
+    },
+    {
+      id: 2,
+      name: 'Product 2',
+      version: 3.13,
+      body: 'This is a description of the product.',
+    },
+    {
+      id: 3,
+      name: 'Product 3',
+      version: 3.13,
+      body: 'This is a description of the product.',
+    },
+  ];
+  const products = fakeData.map(item => <Product item={item} key={item.id} />);
+  return <div>{products}</div>;
+};
+
+export default Products;
--- a/src/js/components/products.js
+++ b/src/js/components/products.js
@@ -3,55 +3,62 @@
 import * as React from 'react';
 import Avatar from '@salesforce/design-system-react/components/avatar';
 import Card from '@salesforce/design-system-react/components/card';
+import DocumentTitle from 'react-document-title';
 import { Link } from 'react-router-dom';
 
 type ProductType = {
   +id: number,
-  +name: string,
-  +version: number,
-  +body: string,
+  +title: string,
+  +version: string,
+  +description: string,
 };
 
-const Product = ({ item }: { item: ProductType }) => (
+const ProductItem = ({ item }: { item: ProductType }) => (
   <Link
     to={`/products/${item.id}`}
     className="slds-text-link_reset slds-p-around_medium"
   >
     <Card
-      heading={item.name}
-      icon={<Avatar variant="entity" label={item.name} />}
+      heading={item.title}
+      icon={<Avatar variant="entity" label={item.title} />}
     >
       <div className="slds-card__body_inner">
         <div className="slds-text-title">Version {item.version}</div>
-        <p>{item.body}</p>
+        <p>{item.description}</p>
       </div>
     </Card>
   </Link>
 );
 
-const Products = () => {
+const ProductsList = () => {
   const fakeData = [
     {
       id: 1,
-      name: 'Product 1',
-      version: 3.13,
-      body: 'This is a description of the product.',
+      title: 'Product 1',
+      version: '3.130',
+      description: 'This is a description of the product.',
     },
     {
       id: 2,
-      name: 'Product 2',
-      version: 3.13,
-      body: 'This is a description of the product.',
+      title: 'Product 2',
+      version: '4.0.2',
+      description: 'This is a description of the product.',
     },
     {
       id: 3,
-      name: 'Product 3',
-      version: 3.13,
-      body: 'This is a description of the product.',
+      title: 'Product 3',
+      version: '0.1.0',
+      description: 'This is a description of the product.',
     },
   ];
-  const products = fakeData.map(item => <Product item={item} key={item.id} />);
-  return <div>{products}</div>;
+  const products = fakeData.map(item => (
+    <ProductItem item={item} key={item.id} />
+  ));
+  return (
+    <DocumentTitle title="Products | MetaDeploy">
+      <div>{products}</div>
+    </DocumentTitle>
+  );
 };
 
-export default Products;
+export default ProductsList;
--- a/src/sass/components/_header.scss
+++ b/src/sass/components/_header.scss
@@ -6,3 +6,7 @@
   font-weight: $font-weight-regular;
   text-transform: uppercase;
 }
+
+.global-nav {
+  text-transform: uppercase;
+}
--- a/flow-typed/design-system-react.js
+++ b/flow-typed/design-system-react.js
@@ -33,3 +33,9 @@ declare module '@salesforce/design-system-react/components/page-header' {
 
   declare export default Component;
 }
+
+declare module '@salesforce/design-system-react/components/card' {
+  import typeof { Component } from 'react';
+
+  declare export default Component;
+}
--- a/src/js/components/404.js
+++ b/src/js/components/404.js
@@ -6,10 +6,7 @@ import { Link } from 'react-router-dom';
 
 const FourOhFour = () => (
   <DocumentTitle title="404 | MetaDeploy">
-    <div
-      className="slds-grow slds-shrink-none slds-text-longform
-        slds-p-horizontal_medium slds-p-vertical_large"
-    >
+    <div>
       <h1 className="slds-text-heading_large">Oh Noes!</h1>
       <p>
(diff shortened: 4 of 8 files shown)


metadeploy/api/models.py
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+
+
+class Project(models.Model):
```
jgerigmeyer: The data model doc and UI both use "Product" here -- is there a reason for the API to use "Project"? Seems like an early recipe for terminology confusion...

wlonk: Per in-person conversation, the existing data model doc uses both, but we'll standardize to "Product".

wlonk:
IMPORTANT! If you've already run migrations on this branch, don't just run them again. Run `python manage.py migrate api zero` BEFORE applying 0302493 Then, check out this commit and run `python manage.py migrate`.

jgerigmeyer:
@wlonk Two things that should happen at some point on this branch:

- script to populate sample products?
- pytest coverage

wlonk:
I'm in the middle of working on 2 right now. I'll add in 1.

metadeploy/api/models.py
```diff
@@ -2,6 +2,16 @@
 
 
 class Product(models.Model):
+    CATEGORY_CHOICES = (
+        ('sfdo', "SFDO"),
+        ('community', "Community"),
```
jgerigmeyer: Can we make whatever gets serialized via the API a human-readable string? That way the front-end doesn't need to hardcode any logic, and can just trust the categories given from the API. E.g. `Salesforce` and `Community` for now.

CONTRIBUTING.rst
```diff
@@ -108,6 +108,11 @@ The WebSocket server (to live-reload on file changes) listens on
 Logging in with Salesforce
 --------------------------
 
+To setup the Salesforce OAuth integration, run the
+``populate_social_apps`` management command::
+
+   python manage.py populate_social_apps --prod-id XXX --prod-secret YYY
+
```
wlonk: Did you duplicate this part of the docs?

src/js/components/products.js
```diff
@@ -0,0 +1,116 @@
...
+import { fetchProducts } from 'products/actions';
+
+import type {
+  Products as ProductsType,
+  Product as ProductType,
+} from 'products/reducer';
+
+type ProductsMapType = Map<string, Array<ProductType>>;
+
+const selectProductsState = (appState): ProductsType => appState.products;
```
wlonk: Is there an implicit `{}` around this? What's this `:` syntax about?

src/js/components/products.js
```diff
@@ -0,0 +1,116 @@
...
+  productsByCategory: selectProductsByCategory(appState),
+});
+
+const actions = {
+  doFetchProducts: fetchProducts,
+};
+
+const ProductItem = ({ item }: { item: ProductType }) => (
+  <Link
+    to={`/products/${item.id}`}
```
wlonk: Does the React router provide a way to do this by a name, not by string interpolation?

src/js/components/products.js
```diff
@@ -0,0 +1,116 @@
...
+    </Card>
+  </Link>
+);
+
+class ProductsList extends React.Component<{
+  productsByCategory: ProductsMapType,
+  doFetchProducts: typeof fetchProducts,
+}> {
+  componentDidMount() {
+    // @@@ Do we need to do this every time?
```
wlonk: Any answer to this question becomes "let's be clear on when where why and how long we cache".

jgerigmeyer:
Note that much like 0302493 above... if you've already run the api "0002_product_category" migration on this branch, don't just run them again. Run `python manage.py migrate api 0001` BEFORE applying 0b6ce7b. Then, check out this commit and run `python manage.py migrate`.

CONTRIBUTING.rst
jgerigmeyer: Not intentionally... probably a merge/rebase error? 🤷‍♂️

src/js/components/products.js
jgerigmeyer: This is equivalent to:

```js
const selectProductsState = appState => appState.products;
```

Which is roughly equivalent to:

```js
const selectProductsState = function(appState) {
  return appState.products;
};
```

I think what's throwing you off is the type-checking Flow syntax -- the `:` syntax is the type definition for the fn return value: https://flow.org/en/docs/types/functions/#toc-function-returns

src/js/components/products.js
jgerigmeyer: Can you say more? I'm not sure exactly what you mean "by a name".

src/js/components/products.js
wlonk: Ooooooh type annotations, I see it now.

src/js/components/products.js
wlonk: Like we're doing with the API URLs, like `URLs.product_instance(id)` to make up an example.

src/js/components/products.js
jgerigmeyer: Yep. I'm already using a front-end caching library (wrapper over IndexedDB) which makes it very easy to add a "max age" after which the cache is considered stale and not used. So I think there would be a number of options:

1. Fetch once per app-load.
2. Fetch once per product-list route-load.
3. Only fetch if the cache is stale (hasn't been fetched in `x` days/hours/minutes).

wlonk: My inclination then is 2, until we have a reason to move to 3.

src/js/components/products.js
jgerigmeyer: Oh, sure. It would be a front-end-only abstraction for convenience, but that makes sense. I'll add it.

src/js/components/products.js
jgerigmeyer: Yep; that's what it's doing now. Just wanted to note the question so we can reconsider once more of the app is in place.

src/js/components/products.js
wlonk: The reason I want it is that it becomes a checkable symbol; a typo in a string can go unnoticed longer than a typo in a symbol.

src/js/components/products.js
wlonk: Cool, and now it's in my head too 😄

jgerigmeyer:
I made some changes to 0003 and 0004 migrations in 01307ec. If you've already run those migrations locally: `python manage.py migrate api 0002`, then check out this commit and run `python manage.py migrate`.

@wlonk I'd appreciate your review, especially on the Python changes I made.

metadeploy/api/management/commands/populate_sample_products.py
```diff
@@ -7,24 +7,46 @@ class Command(BaseCommand):
...
+            category='salesforce',
+            color='#c23934',
+        )
+        Product.objects.create(
+            title=f'Custom Icon Salesforce Product',
+            description=f'This product should have a custom icon.',
+            version='0.1.0',
+            category='salesforce',
+            icon_url='https://lightningdesignsystem.com/assets/images/'
+            'avatar3.jpg',
```
wlonk: Is our linter OK with this? Huh!

metadeploy/api/models.py
```diff
@@ -51,4 +51,4 @@ def icon(self):
                 'category': self.slds_icon_category,
                 'name': self.slds_icon_name,
             }
-        return {'type': 'none'}
+        return None
```
wlonk: Do you really want it to be `None` then? I assumed that just running a check on the `.icon.type` would be easier than null-checking.

metadeploy/api/management/commands/populate_sample_products.py
jgerigmeyer: Yeah, kinda ugly... I fixed it in the migration, but figured I didn't care as much here.

metadeploy/api/models.py
jgerigmeyer: I guess I could go either way, but this seemed cleaner and more descriptive to me.

.flowconfig
```diff
@@ -8,6 +8,7 @@
 
 [lints]
 all=warn
+sketchy-null-string=off
```
mirisuzanne: sooo sketchy ;)

src/sass/components/_header.scss
```diff
@@ -1,8 +1,16 @@
-// Footer
+// Header
 // ======
 
+.page-header {
+  background: $color-background-backdrop-tint;
```
mirisuzanne: Is it important to override this?

src/sass/components/_header.scss
```diff
@@ -1,8 +1,16 @@
...
+}
+
 [data-logo-bit='end'] {
   color: $color-text-brand;
   font-weight: $font-weight-regular;
   text-transform: uppercase;
 }
+
+.global-nav {
+  text-transform: uppercase;
```
mirisuzanne: Also not sure if this is needed?

.flowconfig
jgerigmeyer: I don't want to hear about those sketchy strings! 🙉

jgerigmeyer:
@SondraE The specific style questions @mirisuzanne is referring to relate to 1) whether the "products" nav in the header is all-caps or not, and 2) the background color of the header.

screenshot 2018-08-29 15 08 09
screenshot 2018-08-29 15 07 24
screenshot 2018-08-29 15 06 44

src/sass/components/_products.scss
```diff
@@ -0,0 +1,8 @@
+// Products
+// ========
+
+.slds-icon-standard-account {
+  .has-custom-color & {
+    background-color: var(--custom-color);
+  }
+}
```
jgerigmeyer: @mirisuzanne I wanted to be sure you saw this (and the corresponding markup above: https://github.com/SFDO-Tooling/metadeploy/pull/17/files#diff-e2e03c7c275bd13755ea557c8e6bee82R85). I'm not sure if this is a good idea or not... but unfortunately without duplicating quite a bit of react-SLDS logic (or submitting a PR) we don't have a good way to change the default `background-color` on an avatar, nor to add a class to the avatar itself. I thought this pattern might be okay?

mirisuzanne: That's a good solution. I pushed a slightly simplified version.

SondraE:
@jgerigmeyer 
**1) whether the "products" nav in the header is all-caps or not**
We are making up this nav style, so it doesn't need to be all-caps. Let's make the first letter uppercase following that pattern from [Global Navigation](https://www.lightningdesignsystem.com/components/global-navigation/)

**2) the background color of the header**
I had styled the header bar similar to the [Global Header](https://www.lightningdesignsystem.com/components/global-header/) which has a white background with a shadow. I was thinking we may want to use the [Page Header](https://www.lightningdesignsystem.com/components/page-headers/) for the names and descriptions of the products. Since Page Header is gray, I thought the white Global Header would be a good contrast. You can kinda see how those two elements fit together on this page (https://www.lightningdesignsystem.com/guidelines/messaging/components/illustration-inline-text/). Of course, we aren't using the Global Navigation element that appears in between them. That was my reasoning, but I'm open to doing it differently.

jgerigmeyer:
1) Okay, I'll make that change.
2) Makes sense to me. I'll leave it white for now, and we can always change it later.

PR #19, opened 2018-08-27
[starting-data] 📦 Add management commands for social apps and superusers

wlonk:
elephant

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -108,10 +108,6 @@ The WebSocket server (to live-reload on file changes) listens on
 Logging in with Salesforce
 --------------------------
 
-To make an initial local user::
-
-   python manage.py createsuperuser
-
 To setup the Salesforce OAuth integration, run the
 ``populate_social_apps`` management command::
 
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.rst
```diff
@@ -112,19 +112,27 @@ To make an initial local user::
...
 shared Keybase team folder (``metadeploy/prod.db``).
 
 If you don't have a Salesforce account, ask `Kit <mailto:kit@oddbird.net>`_ to
 send you an invitation by email.
 
+Once you've logged in, you probably want to make your user a superuser.
+You can do that easily via the ``promote_superuser`` management
+command::
+
+   python manage promote_superuser <your email>
```
jgerigmeyer: This should replace the `createsuperuser` instructions above, yes?

requirements/test.txt
```diff
@@ -6,3 +6,5 @@ pytest==3.7.2
 pytest-cov==2.5.1
 pytest-django==3.4.2
 flake8==3.5.0
+factory_boy
+pytest-factoryboy
```
jgerigmeyer: Should these be pinned?

wlonk: Honestly, I'm ambivalent about pinning local/test dependencies, but as long as we're pinning them elsewhere, we should pin them here too.

PR #24, opened 2018-08-30
[custom-domain] Allow login via custom domain.

jgerigmeyer:
https://trello.com/c/eNVFoiGx

puffin


jgerigmeyer:
I think this is functional, but I can't really test until I have a custom domain to try against.

@mirisuzanne @stacyk @SondraE I'm not sure about the styles, since we ended up going with links instead of buttons. This is what it currently looks like:

screenshot 2018-08-30 16 25 55

At the very least it needs something so that it doesn't change width while you're typing in the text input.

The markup lives in `header.js`, the `CustomDomainForm` component.

mirisuzanne:
That looks good to me. I can take a look at the width-issue tomorrow if no one else gets to it first.

jgerigmeyer:
@mirisuzanne I think this is just waiting on final review from you?

PR #27, opened 2018-09-03
Add aria-describedby to custom-domain-login.

jgerigmeyer:
Overlooked commit from #24.

### DIFF
--- a/src/js/components/header.js
+++ b/src/js/components/header.js
@@ -73,8 +73,14 @@ class CustomDomainForm extends React.Component<{}, { url: string }> {
           onChange={this.handleChange}
           assistiveText={null}
           disabled={!window.api_urls.salesforce_custom_login}
+          aria-describedby="login-custom-domain-help"
         >
-          <p className="slds-p-vertical_x-small slds-truncate">
+          <p
+            id="login-custom-domain-help"
+            className="slds-form-element__help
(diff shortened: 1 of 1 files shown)


jgerigmeyer:
@mirisuzanne I meant to get this into #24 but forgot... Feel free to review and merge.

jgerigmeyer:
@mirisuzanne (cc @stacyk @SondraE)

First, I did test with VoiceOver on FF and was able to navigate the menu fine (after I re-learned all the keyboard shortcuts for using VO), including the custom domain form.

I really don't know what to do about keyboard accessibility for the custom-domain form though. The react-SLDS `Dropdown` component is built in such a way that it's impossible for any custom content (outside the standard menu options) to be keyboard accessible -- it hijacks tab to close the menu if it's open, and there's no option to override that. So I think our options are something like:

1. Oh well.
2. Work around react-SLDS, using custom markup and JS logic.
3. Use the standard menu for all three options, moving the actual custom-domain form into a modal or standalone page (shown after the menu option is selected).
4. File a bug with react-SLDS, and submit a PR that makes custom content available.

jgerigmeyer:
I added a new Trello card with this conversation, so it can be discussed and prioritized there: https://trello.com/c/3B5KWFIU

PR #30, opened 2018-09-05
[start-install] Start install on SF organization

wlonk:
This is rough, but it starts the process.

You'll need to set a few env variables:

```
CONNECTED_APP_CLIENT_SECRET
CONNECTED_APP_CALLBACK_URL
CONNECTED_APP_CLIENT_ID
GITHUB_TOKEN
```

Eastern Black Rhino

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -106,6 +106,7 @@ INSTALLED_APPS = [
     'django_extensions',
     'whitenoise.runserver_nostatic',
     'django.contrib.staticfiles',
+    'django_rq',
     'allauth',
     'allauth.account',
     'allauth.socialaccount',
@@ -258,6 +259,17 @@ ACCOUNT_UNIQUE_EMAIL = False
 
 JS_REVERSE_JS_VAR_NAME = 'api_urls'
 JS_REVERSE_EXCLUDE_NAMESPACES = ['admin']
+
+
+RQ_QUEUES = {
+    'default': {
+        'HOST': 'localhost',
+        'PORT': 6379,
+        'DB': 0,
+        'PASSWORD': '',
+        'DEFAULT_TIMEOUT': 360,
+    },
+}
 
 
 # Raven / Sentry
--- a/metadeploy/api/jobs.py
+++ b/metadeploy/api/jobs.py
@@ -1,0 +1,90 @@
+import os
+import sys
+import contextlib
+from tempfile import TemporaryDirectory
+
+from git import Repo
+
+from cumulusci.core import (
+    config,
+    flows,
+    keychain,
+)
+
+from django_rq import job
+
+from django.contrib.auth import get_user_model
+User = get_user_model()
+
+
+@contextlib.contextmanager
+def cd(path):
+    prev_cwd = os.getcwd()
+    os.chdir(path)
+    try:
+        yield
+    finally:
+        os.chdir(prev_cwd)
+
+
+@job
+def run_flow_job(*args, **kwargs):  # pragma: nocover
+    return run_flow(*args, **kwargs)
+
+
+def run_flow(token, instance_url, package_url, flow_name):
+    with TemporaryDirectory() as tmpdirname:
+        with cd(tmpdirname):
+            # Get cwd into Python path, so that the tasks below can
+            # import from the checked-out repo:
+            sys.path.insert(0, '')
+
+            # Let's clone the repo locally:
+            Repo.clone_from(package_url, tmpdirname)
+
+            # There's a lot of setup to make configs and keychains, link
+            # them properly, and then eventually pass them into a flow,
+            # which we then run:
+            org_config = config.OrgConfig({
+                'access_token': token.token,
+                'instance_url': instance_url,
+                'refresh_token': token.token_secret,
+            }, 'test')
+            proj_config = config.YamlProjectConfig(config.YamlGlobalConfig())
+            proj_keychain = keychain.BaseProjectKeychain(proj_config, None)
+            proj_keychain.set_org(org_config)
+            proj_config.set_keychain(proj_keychain)
+
+            # Set up the connected_app:
+            connected_app = config.ServiceConfig({
+                'client_secret': os.environ['CONNECTED_APP_CLIENT_SECRET'],
+                'callback_url': os.environ['CONNECTED_APP_CALLBACK_URL'],
+                'client_id': os.environ['CONNECTED_APP_CLIENT_ID'],
+            })
+            proj_config.keychain.set_service(
+                'connected_app',
+                connected_app,
+                True,
+            )
+
+            # Set up github:
+            github_app = config.ServiceConfig({
+                # It would be nice to only need the token:
+                'token': os.environ['GITHUB_TOKEN'],
+                'password': os.environ['GITHUB_TOKEN'],
+                'email': 'test@example.com',
+                'username': 'not-a-username',
+            })
+            proj_config.keychain.set_service('github', github_app, True)
+
+            # Make and run the flow:
+            flow_config = proj_config.get_flow(flow_name)
+            flowinstance = flows.BaseFlow(
+                proj_config,
+                flow_config,
+                proj_keychain.get_org('test'),
+                options={},
+                skip=[],
+                name=flow_name,
+            )
+            flowinstance()
--- a/metadeploy/api/serializers.py
+++ b/metadeploy/api/serializers.py
@@ -31,3 +31,9 @@ class ProductSerializer(serializers.ModelSerializer):
             'color',
             'icon',
         )
+
+
+class TriggerInstallSerializer(serializers.Serializer):
+    instance_url = serializers.URLField()
+    package_url = serializers.URLField()
+    flow_name = serializers.CharField()
--- a/metadeploy/api/views.py
+++ b/metadeploy/api/views.py
@@ -1,9 +1,33 @@
-from rest_framework import viewsets
+from rest_framework import viewsets, views, status, permissions
+from rest_framework.response import Response
 
-from .serializers import ProductSerializer
+from . import jobs
+from .serializers import (
+    ProductSerializer,
+    TriggerInstallSerializer,
+)
 from .models import Product
 
 
 class ProductViewSet(viewsets.ModelViewSet):
     serializer_class = ProductSerializer
     queryset = Product.objects.all()
+
+
+class TriggerBuildView(views.APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def post(self, request):
+        serializer = TriggerInstallSerializer(data=request.data)
+        if serializer.is_valid():
+            # TODO: this may not be an accurate token?
+            user = request.user
+            token = user.socialaccount_set.first().socialtoken_set.first()
+            jobs.run_flow_job.delay(
+                token,
+                serializer.validated_data['instance_url'],
+                serializer.validated_data['package_url'],
+                serializer.validated_data['flow_name'],
+            )
+            return Response('', status=status.HTTP_202_ACCEPTED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
--- a/metadeploy/urls.py
+++ b/metadeploy/urls.py
@@ -22,6 +22,7 @@ urlpatterns = [
     url(r'^admin/', admin.site.urls),
     url(r'^accounts/', include('allauth.urls')),
     url(r'^api/', include('metadeploy.api.urls')),
+    url(r'^django-rq/', include('django_rq.urls')),
     # Catchall for the rest:
     url(
         r'^(?!admin|accounts|api)',
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
 web: gunicorn metadeploy.wsgi:application --preload
+worker: python manage.py rqworker default
--- a/metadeploy/api/tests/jobs.py
+++ b/metadeploy/api/tests/jobs.py
@@ -1,0 +1,29 @@
+from collections import namedtuple
+
+from ..jobs import run_flow
+
+Token = namedtuple('Token', ('token', 'token_secret'))
+
+
+def test_run_flow(mocker):
+    # TODO: I don't like this test at all. But there's a lot of IO that
+    # this code causes, so I'm mocking it out.
+    mocker.patch('git.Repo.clone_from')
+    mocker.patch('cumulusci.core.config.OrgConfig')
+    mocker.patch('cumulusci.core.config.ServiceConfig')
+    mocker.patch('cumulusci.core.config.YamlGlobalConfig')
+    mocker.patch('cumulusci.core.config.YamlProjectConfig')
+    mocker.patch('cumulusci.core.keychain.BaseProjectKeychain')
+    base_flow = mocker.patch('cumulusci.core.flows.BaseFlow')
+
+    token = Token('token', 'token_secret')
+    instance_url = 'https://example.com/'
+    package_url = 'https://example.com/'
+    flow_name = 'test_flow'
+
+    run_flow(token, instance_url, package_url, flow_name)
+
+    # TODO assert? What we really need to assert is a change in the SF
+    # org, but that'd be an integration test.
+
+    assert base_flow.called
--- a/metadeploy/api/tests/templatetags_api_bootstrap.py
+++ b/metadeploy/api/tests/templatetags_api_bootstrap.py
@@ -4,7 +4,11 @@ from ..templatetags.api_bootstrap import serialize
 
 
 @pytest.mark.django_db
-def test_serialize(user):
+def test_serialize(user_factory):
+    user = user_factory(
+        email='template_tags@example.com',
+        username='template_tags@example.com',
+    )
     actual = serialize(user)
-    expected = "{&quot;username&quot;: &quot;&quot;, &quot;email&quot;: &quot;user_1@example.com&quot;}"  # noqa
+    expected = "{&quot;username&quot;: &quot;template_tags@example.com&quot;, &quot;email&quot;: &quot;template_tags@example.com&quot;}"  # noqa
     assert actual == expected
--- a/metadeploy/api/tests/views.py
+++ b/metadeploy/api/tests/views.py
@@ -1,0 +1,31 @@
+import pytest
+
+
+@pytest.mark.django_db
+def test_trigger_build_view__good(client, mocker):
+    # This is pretty real test data, so we should mock out the resultant
+    # job so it doesn't do IO:
+    run_flow_job = mocker.patch('metadeploy.api.jobs.run_flow_job')
+    data = {
+        'instance_url': 'https://na53.salesforce.com/',
+        'package_url': 'https://github.com/SalesforceFoundation/HEDAP.git',
+        'flow_name': 'install_prod',
+    }
+    response = client.post('/api/start/', data=data)
+
+    assert response.status_code == 202
+    assert run_flow_job.delay.called
+
+
+@pytest.mark.django_db
+def test_trigger_build_view__bad(client, mocker):
+    # This is bad test data, but we should still mock out the resultant
+    # job so it doesn't do IO:
+    mocker.patch('metadeploy.api.jobs.run_flow_job')
+    data = {
+        'package_url': 'https://github.com/SalesforceFoundation/HEDAP.git',
+        'flow_name': 'install_prod',
+    }
+    response = client.post('/api/start/', data=data)
+
+    assert response.status_code == 400
--- a/metadeploy/api/urls.py
+++ b/metadeploy/api/urls.py
@@ -1,8 +1,11 @@
+from django.conf.urls import url
 from rest_framework import routers
 
-from .views import ProductViewSet
+from .views import ProductViewSet, TriggerBuildView
 
 
 router = routers.DefaultRouter()
 router.register(r'products', ProductViewSet)
-urlpatterns = router.urls
+urlpatterns = router.urls + [
+    url(r'^start/$', TriggerBuildView.as_view()),
+]
--- a/metadeploy/conftest.py
+++ b/metadeploy/conftest.py
@@ -1,17 +1,69 @@
+import pytest
+
 import factory
 from pytest_factoryboy import register
 
 from django.contrib.auth import get_user_model
 
+from rest_framework.test import APIClient
+
+from allauth.socialaccount.models import (
+    SocialApp,
+    SocialAccount,
+    SocialToken,
+)
+
 User = get_user_model()
 
 
+@register
+class SocialAppFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = SocialApp
+
+    name = 'Salesforce Production'
+    provider = 'salesforce-production'
+    key = 'https://login.salesforce.com/'
+
+
+@register
+class SocialAccountFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = SocialAccount
+
+    provider = 'salesforce-production'
+
+
+@register
+class SocialTokenFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = SocialToken
+
+    token = factory.Sequence('0123456789abcdef{}'.format)
+    token_secret = factory.Sequence('secret.0123456789abcdef{}'.format)
+    app = factory.SubFactory(SocialAppFactory)
+    account = factory.SubFactory(SocialAccountFactory)
+
+
+@register
 class UserFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = User
 
     email = factory.Sequence("user_{}@example.com".format)
+    username = factory.Sequence("user_{}@example.com".format)
     password = factory.PostGenerationMethodCall('set_password', 'foobar')
+    socialaccount_set = factory.RelatedFactory(SocialAccountFactory, 'user')
 
 
-register(UserFactory)
+@pytest.fixture
+def client(user_factory):
+    user = user_factory()
+    client = APIClient()
+    client.force_login(user)
+    return client
+
+
+# @pytest.fixture
+# def anon_client():
+#     return APIClient()
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,3 +11,5 @@ djangorestframework==3.8.2
 django-colorfield==0.1.15
 raven==6.9.0
 git+https://github.com/SalesforceFoundation/Cumulusci.git@feature/python3
+GitPython==2.1.11
+django-rq==1.2.0
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,3 +10,4 @@ django-js-reverse==0.8.2
 djangorestframework==3.8.2
 django-colorfield==0.1.15
 raven==6.9.0
+git+https://github.com/SalesforceFoundation/Cumulusci.git@feature/python3
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -8,3 +8,4 @@ pytest-django==3.4.2
 flake8==3.5.0
 factory_boy==2.11.1
 pytest-factoryboy==2.0.1
+pytest-mock==1.10.0


metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+    finally:
+        os.chdir(prev_cwd)
+
+
+@job
+def run_flow_job(*args, **kwargs):  # pragma: nocover
+    return run_flow(*args, **kwargs)
+
+
+def run_flow(token, instance_url, package_url, flow_name):
```
wlonk: This function is where the juice is, and what requires especial discussion and consideration.

metadeploy/urls.py
```diff
@@ -22,6 +22,7 @@
     url(r'^admin/', admin.site.urls),
     url(r'^accounts/', include('allauth.urls')),
     url(r'^api/', include('metadeploy.api.urls')),
+    url(r'^django-rq/', include('django_rq.urls')),
```
cdcarter: sooner rather than later we should add a base setting `ADMIN_URL` that's admin for local running but sourced from "DJANGO_ADMIN_URL" in the production settings module.

    # Django Admin, use {% url 'admin:index' %}
    path(settings.ADMIN_URL, admin.site.urls),
    path(
        urllib.parse.urljoin(settings.ADMIN_URL, "django-rq/"),
        include("django_rq.urls"),
    ),

config/settings/base.py
```diff
@@ -260,6 +261,17 @@ def env(name, default=NoDefaultValue, type_=str):
 JS_REVERSE_EXCLUDE_NAMESPACES = ['admin']
 
 
+RQ_QUEUES = {
```
cdcarter: lets get this a little 12 factor-y... 

use this incantation (well documented setting) to pool the connections used by django cache:
https://github.com/rq/django-rq#support-for-django-redis-and-django-redis-cache

and then something like this to get the redis URL from the env
```
REDIS_LOCATION = "{0}/{1}".format(env("REDIS_URL", default="redis://127.0.0.1:6379"), 0)
CACHES = {
    "default": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": REDIS_LOCATION,
        "OPTIONS": {
            "CLIENT_CLASS": "django_redis.client.DefaultClient",
            # Mimicing memcache behavior.
            # http://niwinz.github.io/django-redis/latest/#_memcached_exceptions_behavior
            "IGNORE_EXCEPTIONS": True,
        },
    }
}
```

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+    with TemporaryDirectory() as tmpdirname:
+        with cd(tmpdirname):
```
cdcarter: nested context managers? time for an [ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack)

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+            # Get cwd into Python path, so that the tasks below can
+            # import from the checked-out repo:
+            sys.path.insert(0, '')
```
cdcarter: this works, but can we `path.abspath(tmpdirname)` instead of the empty string implying cwd?

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+
+            # Let's clone the repo locally:
+            Repo.clone_from(package_url, tmpdirname)
```
cdcarter: okay here's where i'm getting really nitty gritty...

Repo isn't a type that we own and it's not a standard CCI/Django/DRF type, so i don't like it being a naked namespace. lets `import git` and use `git.Repo`

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+            flow_config = proj_config.get_flow(flow_name)
+            flowinstance = flows.BaseFlow(
+                proj_config,
+                flow_config,
+                proj_keychain.get_org('test'),
+                options={},
+                skip=[],
+                name=flow_name,
+            )
+            flowinstance()
```
cdcarter: totally fine with ignoring what the log output of the flow is, but i absolutely want appserver log lines before and after flow execution.

cdcarter: also, lets `return flowinstance.step_return_values` here

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+                'refresh_token': token.token_secret,
+            }, 'test')
+            proj_config = config.YamlProjectConfig(config.YamlGlobalConfig())
+            proj_keychain = keychain.BaseProjectKeychain(proj_config, None)
+            proj_keychain.set_org(org_config)
+            proj_config.set_keychain(proj_keychain)
+
+            # Set up the connected_app:
+            connected_app = config.ServiceConfig({
+                'client_secret': os.environ['CONNECTED_APP_CLIENT_SECRET'],
```
cdcarter: lets pull these (env var based values) into the settings module and have them default to the env var.

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+            proj_config.keychain.set_service(
+                'connected_app',
+                connected_app,
+                True,
+            )
+
+            # Set up github:
+            github_app = config.ServiceConfig({
+                # It would be nice to only need the token:
+                'token': os.environ['GITHUB_TOKEN'],
```
cdcarter: same thing here, get from django settings.

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+                'password': os.environ['GITHUB_TOKEN'],
+                'email': 'test@example.com',
+                'username': 'not-a-username',
```
cdcarter: aaand even these constants

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+            })
+            proj_config.keychain.set_service('github', github_app, True)
+
+            # Make and run the flow:
+            flow_config = proj_config.get_flow(flow_name)
+            flowinstance = flows.BaseFlow(
+                proj_config,
+                flow_config,
+                proj_keychain.get_org('test'),
```
cdcarter: store the org in keychain as 'current_org' or something else more descriptive

metadeploy/api/views.py
```diff
@@ -1,9 +1,33 @@
...
     queryset = Product.objects.all()
+
+
+class TriggerBuildView(views.APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def post(self, request):
+        serializer = TriggerInstallSerializer(data=request.data)
+        if serializer.is_valid():
+            # TODO: this may not be an accurate token?
```
cdcarter: its definitely not always, but im okay with it being a todo for now.

metadeploy/api/views.py
```diff
@@ -1,9 +1,33 @@
...
+            user = request.user
+            token = user.socialaccount_set.first().socialtoken_set.first()
+            jobs.run_flow_job.delay(
+                token,
+                serializer.validated_data['instance_url'],
+                serializer.validated_data['package_url'],
+                serializer.validated_data['flow_name'],
```
cdcarter: lets be ready to have to pass request.user here

wlonk: Yeah. I think what we're ultimately gonna do is just create a model here (something like a Build, but I think we have a different name in the data model document), and trigger a job that will look up all the relevant data off that model.

metadeploy/api/jobs.py
cdcarter: I mean, it's definitely not quite python-module-namespace-semantics but it helps the brain know that we're dealing with a `git.Repo` from a third party package. Meh I won't clock you too frequently for it.

metadeploy/api/jobs.py
cdcarter: the github checkout *is* happening in the context of a user, in the case of our private repos, so it's not TOTALLY crazy. but. yea, the nature of a Services required parts is a little weird. sounds good to come back to.

wlonk: We absolutely need a token, though, as without that we can't guarantee completion of the login step—if the user has 2FA enabled it'll expect that and fail. So requiring an alternative set of settings which aren't used anyway when there's a token seems wrong.

metadeploy/api/jobs.py
wlonk: If code is read more often than it's written (and that's the core insight that led to Python, right?) then adhering to the stylistic norms of the community that'll be reading it is important and part of the goal, so I'm happy to take this approach!

metadeploy/api/jobs.py
cdcarter: acknowledged. https://github.com/SalesforceFoundation/CumulusCI/issues/728

metadeploy/api/views.py
cdcarter: yep, makes a ton of sense. we can talk more when you get there but....

we did that for MetaCI at first, but were running into issues where when multiple workers were already scaled up, the build would get dispatched before the transaction finished.  (yea yea i know this isn't a novel problem to have with a job queue...). In MetaCI, the current implementation is to queue a job on a `short` queue that checks the build is ready to go, and then re-enqueues it onto `default`.

personally, I'd like us to move more toward something like https://brandur.org/job-drain where we put the job into postgres inside a transaction, and then drain it to the main rq.

@jlantz - have you put thought to if we want to just pay for hirefire for metadeploy for worker scaling or if we want to solution for that?

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,102 @@
...
+        flowinstance()
+
+
+# TODO:
+#
+# We'll want to subclass BaseFlow and add logic in the progress
+# callbacks to record and possibly push progress:
+# pre_flow, post_flow, pre_task, post_task, pre_subflow, post_subflow
+#
+# Can we do anything meaningful with a return value from a @job, too?
```
cdcarter: job return values ARE popped back into redis as a result, and also shown in the rq admin ui http://python-rq.org/docs/results/

wlonk: Sorry, I mean, like, is there a value we can return here that's useful in that context? The flow executes synchronously, but is there information in it that is useful beyond what we can get in the progress hooks?

config/settings/base.py
```diff
@@ -260,6 +265,23 @@ def env(name, default=NoDefaultValue, type_=str):
...
+        'DB': 0,
+        'PASSWORD': '',
+        'DEFAULT_TIMEOUT': 360,
+    },
+}
+
+
+CONNECTED_APP_CLIENT_SECRET = env('CONNECTED_APP_CLIENT_SECRET')
+CONNECTED_APP_CALLBACK_URL = env('CONNECTED_APP_CALLBACK_URL')
+CONNECTED_APP_CLIENT_ID = env('CONNECTED_APP_CLIENT_ID')
```
cdcarter: Oh! I remember what's slightly wrong with doing it this way... But I have a proposed solution too. In Heroku, changing an environment variable restarts all dynos. But these env vars change together. We ran into this annoyance in one of our apps where to switch out the connected apps, we end up triggering a bunch of unnecessary restarts. I think maybe we wanna just contain a serialized representation of the ServiceConfig, instead of the constituent parts. This would hold true for github too. this way also, if we change the definition of the service, we just update the CCI version and the env var in one go.

metadeploy/api/jobs.py
cdcarter: ah yea, no, nothing that we couldn't get out of `_post_flow`.

metadeploy/api/views.py
jlantz: I think using Hirefire for now makes sense.  Here's a link to the view that implements the endpoint Hirefire calls to get current queue status: https://github.com/SalesforceFoundation/MetaCI/blob/master/metaci/hirefire/views.py

In the longer run, we'll probably want to integrate whatever solution we decide for natively scaling up worker dynos in MetaCI.  I'm still leaning towards one off dynos for that but we'll cross that bridge when we get there.

config/settings/base.py
wlonk: Interesting. I tend to use the command line to do changes like that, so I wouldn't have thought about it triggering multiple restarts. Is there any character we can be sure that the secret, callback, and client ID _won't_ include, to split on?

cdcarter: interesting, I didn't realize you could set multiple env vars atomically via the command line app for heroku! the web interface does them one at a time. 

I just checked and in one of our other apps we have used the Env.json() method provided by https://pypi.org/project/django-environ/ to deserialize a json blob. django-environ is part of the django-cookiecutter template so it's always "been there" for our other apps. 

that said, since we can change them via the CLI this is less big of a deal. we can icebox using django-environ if its more than a few minutes for you to get acqunited.

metadeploy/api/models.py
```diff
@@ -20,6 +23,8 @@ class Product(models.Model):
     title = models.CharField(max_length=256)
     description = models.TextField()
     version = models.CharField(max_length=256)
+    # TODO: We have reason to want this to be a disinct model, so it can
+    # be set in the Django Admin:
```
cdcarter: it's probably just easiest to make this a model, but we could go low tech and do an environment variable provided json array of category labels - loaded/cached into settings... 🤷‍♂️  i don't think we have any additional category metadata. that said, maybe we just do it now and are thankful later.

metadeploy/api/models.py
```diff
@@ -52,3 +57,27 @@ def icon(self):
                 'name': self.slds_icon_name,
             }
         return None
+
+
+class Job(models.Model):
+    token = models.CharField(max_length=256)
```
cdcarter: is this a UUID?

metadeploy/api/views.py
```diff
@@ -31,3 +35,9 @@ def post(self, request):
             )
             return Response('', status=status.HTTP_202_ACCEPTED)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class JobViewSet(viewsets.ModelViewSet):
+    serializers_class = JobSerializer
+    queryset = Job.objects.all()
```
cdcarter: so, right now any logged in user could see any job. is https://trello.com/c/3Iopq2KN/27-as-an-admin-i-want-a-permalink-to-view-share-the-status-of-the-current-installation-job a sufficient story to ensure that we firm that ACL later?

metadeploy/api/models.py
```diff
@@ -52,3 +57,27 @@ def icon(self):
...
+    instance_url = models.URLField()
+    package_url = models.URLField()
+    flow_name = models.CharField(max_length=64)
+
+    def save(self, *args, **kwargs):
+        instance = super().save(*args, **kwargs)
+        # TODO: We may have to ensure that the save has completed
+        # writing to the database, assuming we exist in a transaction
+        # here, to be guaranteed to kick off the background job:
+        jobs.run_flow_job.delay(
```
cdcarter: here's christian, asking for a logline! i suspect we can configure RQ to log every delay() on the webserver.

metadeploy/api/models.py
cdcarter: oh no, its the bearer token! yea, i'd much rather that live in allauths court, so that we don't have to scrub it out here too. I think the trick here is going to be to just store the user or org id and then lookup the token from the db in the delayed job, and then invalidating it at the end of the job.

metadeploy/api/models.py
cdcarter: specifically we get a job uuid from rq that's very useful to lookup exceptions if a job gets lost

metadeploy/api/models.py
wlonk: Here's my worry about that (and what I have now doesn't address this worry, either, just to be clear about it):

I start a job. I start another. Job 1 finishes before job 2 (which I've enqueued and expect to work) starts. Job 2 now has no valid token.

This is gonna be an issue we need to deal with one way or another.

metadeploy/api/views.py
wlonk: Yes, that story will include both access and not-access considerations.

metadeploy/api/models.py
wlonk: I think I'd prefer a model, even if it's tiny, just because then someone managing an install can manage it like they'd manage anything else via the admin.

metadeploy/api/models.py
cdcarter: yea, I think we need to define a token retention period and sweep them n hours if there are no jobs enqueued connected to that org/user id.

cdcarter: that "sweep" would probably also hit the token revocation endpoint

metadeploy/api/jobs.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def run_flow(token, token_secret, instance_url, package_url, flow_name):
+    with contextlib.ExitStack() as stack:
+        tmpdirname = stack.enter_context(TemporaryDirectory())
+        stack.enter_context(cd(tmpdirname))
+
+        # Get cwd into Python path, so that the tasks below can import
+        # from the checked-out repo:
+        sys.path.insert(0, os.path.abspath(tmpdirname))
```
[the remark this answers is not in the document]
cdcarter: oh thats clever.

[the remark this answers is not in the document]
davisagli: I'm concerned that it may end up being too clever, but worth an experiment anyway.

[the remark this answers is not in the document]
wlonk: > This is a dictionary that maps module names to modules which have already been loaded. This can be manipulated to force reloading of modules and other tricks. However, **replacing the dictionary will not necessarily work as expected** and deleting essential items from the dictionary may cause Python to fail.

Emphasis mine.

That doesn't give me much hope for munging `sys.modules` doing what we hope.

If we don't munge it, then the imports that happen in the install process _will_ modify `sys.modules`, which could lead to other surprises, but I think is sufficiently unlikely to cause problems yet that I'm OK deferring decisions on this.

config/settings/base.py
wlonk: Having looked over django-environ, I am having some aesthetic conflicts. I think I'd rather start with "change via command line" for now, and return to this if there's an issue.

metadeploy/api/views.py
wlonk: I've looked over Brandur's approach, and it makes a lot of sense, and seems to be the cleanest option, though I'd want to be very clear in the docs that it's what we're doing.

So, we'd make a `ScheduledJob` model, and set up a periodic enqueuer task via rq-scheduler that runs every second (say), reads the `ScheduledJob` table, kicks off a task for each with appropriate data from the table, and clears the `ScheduledJob` models (or marks them as enqueued, if we want to keep them for historical purposes).

metadeploy/api/apps.py
```diff
@@ -0,0 +1,19 @@
...
+
+class ApiConfig(AppConfig):
+    name = 'metadeploy.api'
+    verbose_name = 'API'
+
+    def ready(self):
+        from .jobs import run_enqueuer
+
+        scheduler = get_scheduler('default')
+        scheduler.schedule(
```
cdcarter: scheduling this in the appconfig is going to bring pain - https://github.com/rq/rq-scheduler/issues/121

metaci uses [django-rq-scheduler](https://pypi.org/project/django-rq-scheduler/) to do database backed scheduling. not married to it, but it's pretty simple and works. 

otherwise I'd do a management command that starts the scheduler queue itself and then ensures our jobs are scheduled but not duped (mentioned in that SO link)

metadeploy/api/apps.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class ApiConfig(AppConfig):
+    name = 'metadeploy.api'
+    verbose_name = 'API'
+
+    def ready(self):
+        from .jobs import run_enqueuer
+
+        scheduler = get_scheduler('default')
```
cdcarter: we should add a second queue with a low timeout for this very short/simple job. we usually call it ..."short"

metadeploy/api/jobs.py
```diff
@@ -101,13 +128,20 @@ def run_flow(token, token_secret, instance_url, package_url, flow_name):
...
+def enqueuer():
+    for j in Job.objects.filter(enqueued_at=None):
+        run_flow_job.delay(
+            j.token,
+            j.token_secret,
+            j.instance_url,
+            j.package_url,
+            j.flow_name,
+        )
+        j.enqueued_at = timezone.now()
```
cdcarter: can we grab the rq job uuid here and stash it in the model, now that we have code for it?

config/settings/base.py
```diff
@@ -57,7 +57,7 @@ def env(name, default=NoDefaultValue, type_=str):
     except KeyError:
         if default == NoDefaultValue:
             raise ImproperlyConfigured(
-                f"Missing environment variable: {name}."
+                f'Missing environment variable: {name}.'
```
cdcarter: 😮

wlonk: We just tend towards single-quoted strings. Is that alarming?

davisagli: We'll probably start using `black` at some point, and it normalizes to double-quoted. But since it can do that automatically, don't sweat it until then.

cdcarter: it's just always funny to see in a diff, ignore me :)

wlonk: The best testimonial, of course:

> **Carl Meyer**, Django core developer:
>    At least the name is good.

wlonk:
OK, remaining tasks:

 - [ ] Track status via Flow hooks: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215434247
 - [ ] Product categories as a related model: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215777218
 - [ ] Sweep tokens, store User references not tokens etc: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215777740
 - [ ] In another PR, add access controls on jobs: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215778460
 - [ ] Possibly split the job out into smaller testable units: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215984596

Does that sound right? Am I missing anything?

cdcarter:
Which of those things do you want to get into scope for this PR? some thoughts:
- we can start by tracking the overall flow status via flow hooks - figuring out step by step status tracking is going to have to align with how we do optional steps etc and that kinda balooons the size of this
- token sweeping and revocation seems like it might be a bigger story too - maybe start by just switching to passing references and we can come back to the revocation in a soontobehandled item?

leaving the todos for this PR:
 - [ ] Track overall flow status via Flow hooks: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215434247
 - [ ] Product categories as a related model: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215777218
 - [ ] store User references not tokens etc: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215777740
 - [ ] Possibly split the job out into smaller testable units: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215984596

but... yes that list sounds good!

wlonk:
I'm with you; this PR is already at about the maximum size for useful code review, so I think that the bigger and separable things should be separated into their own stories.

I think that storing user references is crucial and I'll do that here, and I'm not sure about splitting the job up; it feels like something I'd like to do, but I don't know how I want to do it yet, so I want to revisit that.

So, here:

 - [ ] Sweep tokens, store User references not tokens etc: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215777740
 - [ ] Product categories as a related model: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215777218

And in new stories:

 - [ ] Track status via Flow hooks: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215434247 (in https://trello.com/c/3Iopq2KN/27-as-an-admin-i-want-a-permalink-to-view-share-the-status-of-the-current-installation-job)
 - [ ] In another PR, add access controls on jobs: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215778460 (also in https://trello.com/c/3Iopq2KN/27-as-an-admin-i-want-a-permalink-to-view-share-the-status-of-the-current-installation-job)
 - [ ] Possibly split the job out into smaller testable units: https://github.com/SFDO-Tooling/metadeploy/pull/30#discussion_r215984596 (in a new icebox ticket, https://trello.com/c/NyEy0FZb/48-as-a-developer-i-want-the-calls-to-the-cumulusci-library-to-be-well-structured-for-testing)

wlonk:
@cdcarter I think this is ready, at least in terms of the punch list being done and other tasks being moved to their own Trello cards. Do you wanna take a look?

metadeploy/api/migrations/0009_migrate_category_to_productcategory.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import migrations
+
+
+def forwards(apps, schema_editor):
+    ProductCategory = apps.get_model('api', 'ProductCategory')
+    Product = apps.get_model('api', 'Product')
+    for product in Product.objects.all():
+        category, _ = ProductCategory.get_or_create(title=product.category)
+        product.category_fk = category
```
cdcarter: should this be setting the fk to the pk instead of the model instance?

metadeploy/api/models.py
```diff
@@ -63,6 +68,18 @@ def icon(self):
         return None
 
 
+class Job(models.Model):
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.PROTECT,
+    )
+    instance_url = models.URLField()
```
cdcarter: doesn't this come from the token too?

metadeploy/api/models.py
```diff
@@ -63,6 +68,18 @@ def icon(self):
...
+    package_url = models.URLField()
```
cdcarter: this is a repo url right? there's such a thing as a package url too, and i sense confusion coming. i mean, ultimately these should be related to the product probably right? but for now, lets call it `repo_url`

metadeploy/api/serializers.py
```diff
@@ -67,3 +68,18 @@ class Meta:
             'image',
             'most_recent_version',
         )
+
+
+class JobSerializer(serializers.ModelSerializer):
+    user = serializers.HiddenField(
+        default=serializers.CurrentUserDefault()
```
cdcarter: http://www.django-rest-framework.org/topics/release-notes/#380
> Previously read_only fields with a default value would use the default for create and update operations.

will this change to drf 3.8.0 be an issue?

metadeploy/api/serializers.py
```diff
@@ -67,3 +68,18 @@ class Meta:
...
+    )
+
+    class Meta:
+        model = Job
+        fields = (
+            'user',
+            'instance_url',
+            'package_url',
```
cdcarter: switch to repo_url

metadeploy/api/migrations/0009_migrate_category_to_productcategory.py
wlonk: Nah; this is just part of a migration from `category` as a CharField to `category` as a `ForeignKey`, using an intermediate field. I just named the intermediate field the first thing I thought of, as we don't keep it around.

metadeploy/api/models.py
wlonk: Oh, yes, I guess it does! I'll make that change.

metadeploy/api/models.py
wlonk: OK, sounds good.

wlonk:
@cdcarter if you give this a 👍 review, I can merge it.

config/settings/base.py
```diff
@@ -267,6 +273,41 @@ def env(name, default=NoDefaultValue, type_=str):
...
+        'DEFAULT_TIMEOUT': 10,
+    },
+}
+
+
+# SF Connected App and GitHub configuration:
+CONNECTED_APP_CLIENT_SECRET = env('CONNECTED_APP_CLIENT_SECRET')
+CONNECTED_APP_CALLBACK_URL = env('CONNECTED_APP_CALLBACK_URL')
+CONNECTED_APP_CLIENT_ID = env('CONNECTED_APP_CLIENT_ID')
+GITHUB_TOKEN = env('GITHUB_TOKEN')
```
jgerigmeyer: @wlonk If these are required, they should be added to `env.example` and possibly also documented in `CONTRIBUTING.rst` (if they cannot be hardcoded in `env.example`).

metadeploy/api/models.py
```diff
@@ -21,10 +26,10 @@ class Product(models.Model):
 
     title = models.CharField(max_length=256)
     description = models.TextField()
-    category = models.CharField(
-        choices=CATEGORY_CHOICES,
-        default='salesforce',
-        max_length=256,
+    category = models.ForeignKey(
+        ProductCategory,
+        null=True,
```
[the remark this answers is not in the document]
wlonk: No, we needed it for a stage in the migration, but it should get cleaned up.

PR #33, opened 2018-09-07
[select-plan] 🛢 Select a plan

jgerigmeyer:
https://trello.com/c/DrUEq32H

squirrel


metadeploy/api/models.py
```diff
@@ -19,13 +19,13 @@ class Product(models.Model):
...
     title = models.CharField(max_length=256)
     description = models.TextField()
-    version = models.CharField(max_length=256)
     category = models.CharField(
         choices=CATEGORY_CHOICES,
         default='salesforce',
         max_length=256,
     )
     color = ColorField(blank=True)
+    image_url = models.URLField(blank=True)
```
cdcarter: this....should probably be a filefield.

metadeploy/api/models.py
```diff
@@ -52,3 +56,48 @@ def icon(self):
                 'name': self.slds_icon_name,
             }
         return None
+
+
+class Version(models.Model):
+    product = models.ForeignKey(Product, on_delete=models.PROTECT)
+    label = models.CharField(max_length=64)
```
cdcarter: lets do a more generous max length, 1000?
--pedantic thoughts--
under the hood a CharField is a `character varying (<max_length>)` column with a TextInput widget. a TextField is a `text` with a `TextArea` widget, and an optional maxlength validator. no matter if the max_length is specified in the DDL, since we're using postgres this is just a varlena under the hood. it used to be that updating the max_length for a varchar in postgres caused table rewriting, but they've fixed that. So... there's no REAL harm in specifying the max_length at the database level, but generally we don't rely on postgres checks. Maybe we should do that more! ...maybe we should write a custom CharField that uses a `text` column but the <input> widget... idrk. but...64 seems small!

metadeploy/api/models.py
```diff
@@ -52,3 +56,48 @@ def icon(self):
...
+    description = models.TextField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    is_production = models.BooleanField(default=True)
+
+    def natural_key(self):
+        return (self.product, self.label)
```
cdcarter: let's uniq_together these columns, since its a true natural key, and feel free to @property it too!

metadeploy/api/models.py
```diff
@@ -52,3 +56,48 @@ def icon(self):
...
+
+    @property
+    def primary_plan(self):
+        # This will raise an error if the number of primary plans != 1:
```
cdcarter: yea, lets get a uniq_together index on these too!!

metadeploy/api/models.py
```diff
@@ -52,3 +56,48 @@ def icon(self):
...
+    def secondary_plan(self):
+        return self.plan_set.filter(tier=Plan.Tier.Secondary).first()
+
+    @property
+    def additional_plans(self):
+        return self.plan_set.filter(tier=Plan.Tier.Additional).order_by('id')
+
+
+class Plan(models.Model):
+    class Tier:
```
cdcarter: we generally use the [Choices()](https://django-model-utils.readthedocs.io/en/latest/utilities.html#choices) object from django_model_utils for enums

metadeploy/api/models.py
cdcarter: though i guess that means materializing Primary

metadeploy/api/models.py
wlonk: 👍  on `unique_together`, but I won't actually make this a `property`, as the Django serialization machinery expects a callable with this name; I am going to add a Manager with a `get_by_natural_key`, though.

metadeploy/api/models.py
wlonk: I'm not quite sure what you're suggesting here. Can you elaborate?

src/sass/components/_products.scss
```diff
@@ -6,3 +6,10 @@
     background-color: var(--custom-color);
   }
 }
+
+.page-header {
+  border: none;
+  border-radius: 0;
+  background-color: $color-background-reminder;
+  box-shadow: none;
```
jgerigmeyer: These are overriding the page-header defaults from SLDS.

src/sass/components/_header.scss
```diff
@@ -1,8 +1,10 @@
 // Header
 // ======
 
-.page-header {
-  background: $color-background-backdrop-tint;
+.global-header {
+  border-radius: 0;
+  background-color: $color-background-backdrop-tint;
+  z-index: 100;
```
jgerigmeyer: @stacyk I'm not 100% sure about this... probably there's a better way? It's allowing the box-shadow from the top page-header (what we're using as a global header) to display over the second page-header.

metadeploy/api/models.py
cdcarter: Whoops, end of day friday brain.

I think what I was getting at (cause...yikes at me...) is that maybe we wanna create a foreignkey on Version back down to primary_plan so the database can hep enforce and optimize queries for the primary Plan.

definitely not super necessary so don't worry. if we decide we wanna do it we can do it later.

metadeploy/api/models.py
```diff
@@ -19,13 +21,13 @@ class Product(models.Model):
...
     title = models.CharField(max_length=256)
     description = models.TextField()
-    version = models.CharField(max_length=256)
     category = models.CharField(
         choices=CATEGORY_CHOICES,
         default='salesforce',
         max_length=256,
     )
     color = ColorField(blank=True)
+    image_url = models.ImageField()
```
jgerigmeyer: This is now broken... we need to configure file uploads.

Also, do we want to rename it if it's a file not a URL?

metadeploy/api/migrations/0008_auto_20180910_1628.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 2.1.1 on 2018-09-10 16:28
```
jgerigmeyer: I find it helpful to give migrations a descriptive filename.

src/sass/components/_products.scss
stacyk: I assume there's no way around having to use the class `slds-page-header` since it is the default in this component? Because we are overriding every CSS property except background-clip. That said, what we are doing works. We could use `unset` to go back to browser defaults but being more explicit makes sense here. I would say since they use `background` instead of `background-color` we may want to edit ours to match (it doesn't make much difference unless there's ever a background-image in play in the future).  It is a bit odd to see a class called "page-header" in the products.scss file. Is this a class we might use somewhere else with these overrides?

src/sass/components/_header.scss
stacyk: I am going to assume this isn't possible, but could the first be the [global header](https://www.lightningdesignsystem.com/components/global-header/) and the second the [page-header](https://www.lightningdesignsystem.com/components/page-headers/)? For the z-index specifically, we could add our z-index mixin for controlling layering in an easy to implement way. Want me to do that?

src/sass/components/_products.scss
jgerigmeyer: No, there's not really a way around the `slds-page-header` class... you're welcome to rename `background-color` if you like, and yes this should be in a more general layout file.

src/sass/components/_header.scss
jgerigmeyer: We started there, but for a number of reasons the page-header made more sense in our context. Please adjust z-index as you see fix.

src/js/utils/routes.js
```diff
@@ -3,7 +3,8 @@
 const routes = {
   home: () => '/',
   product_list: () => '/products',
-  product_detail: (id: number) => `/products/${id}`,
+  product_detail: (id: number | void) =>
+    `/products/${id === undefined ? ':id' : id}`,
```
wlonk: We're gonna do the version detail stuff in another PR I take it?

wlonk:
Apparently adding a commit of my own to address a thing dismissed my review, but see it at https://github.com/SFDO-Tooling/metadeploy/pull/33#pullrequestreview-154402941

cdcarter:
if that gets annoying I can turn our branch protections down a notch since we're pre-prod...

metadeploy/api/models.py
```diff
@@ -38,6 +40,13 @@ class Product(models.Model):
     )
     slds_icon_name = models.CharField(max_length=64, blank=True)
 
+    def __str__(self):
+        return self.title
```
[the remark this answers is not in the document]
jgerigmeyer: Helpful when using the Django admin (which I do frequently).

src/js/utils/routes.js
jgerigmeyer: Yes, that's my plan.

jgerigmeyer:
@cdcarter I don't feel strongly. It's moderately annoying, but in the "I fully understand why this is a good idea" kind of way. 🤷‍♂️

PR #39, opened 2018-09-12
[urls] 📦 🛢 Refactor client-facing URLs to use slugs.

jgerigmeyer:
Also allows viewing a not-most-recent version by directly entering the URL.

https://trello.com/c/MtGuRKJt

mongoose


metadeploy/api/models.py
```diff
@@ -134,3 +199,43 @@ def natural_key(self):
...
+
+
+class PlanSlug(models.Model):
+    """
+    Rather than have a slugfield directly on the Plan model, we have
+    a related model. That way, we can have a bunch of slugs that pertain
+    to a particular model, and even if the slug changes and someone uses
+    an old slug, we can redirect them appropriately.
+    """
+    slug = models.SlugField(unique=True)
```
jgerigmeyer: It looks like `unique=True` is universally enforced... For products we want that to be the case, but for Plans slug uniqueness should only be enforced _within that version_. (This doesn't need to hold up this PR, but will be relevant in the next plan-detail feature.)

src/js/components/404.js
```diff
@@ -12,7 +12,7 @@ const FourOhFour = () => (
       className="slds-text-longform
         slds-p-around_x-large"
     >
-      <h1 className="slds-text-heading_large">Oh Noes!</h1>
+      <h1 className="slds-text-heading_large">Oh No!</h1>
```
cdcarter: fwiw, "Oh Noes" is probably fine voice&tone: https://www.lightningdesignsystem.com/assets/downloads/salesforce-voice-and-tone.pdf?

cdcarter:
that said, if i'm misreading the reacts, please help me understand :)

cdcarter:
I see a story for the (visual) pagination of the product list. my "concern" here is more about the behavior when going directly to a version by URL. if the work makes sense to tackle together, we can just throw a note on card 17, otherwise a new card to make sure we track it through?

cdcarter:
which is to say, not urgent, but I don't want it lost.

metadeploy/api/models.py
```diff
@@ -118,7 +211,35 @@ def additional_plans(self):
...
+        help_text=(
+            'If multiple slugs are active, we will default to the most '
+            'recent.'
+        ),
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        unique_together = (
+            ('slug', 'parent'),
```
wlonk: Wait, is this right? This says you can repeat a slug only if it's on a different Plan. We actually want it to be for that Plan's Version, right? Like `('slug', 'parent__version')`, which I'm not sure will work.

jgerigmeyer:
@cdcarter You're absolutely correct, and I think it should be addressed in this PR. We discussed that exact change, but I wasn't clear on the performance implications (i.e. the # of likely versions). I'll go ahead and adjust that (so that we're querying versions by `product` and `label`, not just `product`).

metadeploy/api/models.py
```diff
@@ -1,20 +1,101 @@
...
+
+    def ensure_slug(self):
+        if not self.slug_queryset.filter(is_active=True).exists():
+            slug = slugify(self.title)
+            slug = self._find_unique_slug(slug)
+            self.slug_class.objects.create(
+                parent=self,
+                slug=slug,
+                is_active=True,
+            )
```
jgerigmeyer: @wlonk Am I right that this `ensure_slug` command should really never be run for a `Plan`, since it doesn't contain the correct logic in `_find_unique_slug` (that is, it will enforce uniqueness globally)?

I guess that command is only used in the tests...

wlonk: Yeah. I think I may have been over-eager in making that method, and we can perhaps remove it for now. Currently, there's logic very much like it in the data migration, but I'm OK with that for now.

jgerigmeyer:
@cdcarter @wlonk I think all the review items have been addressed, if you two have a chance to check over and confirm.

metadeploy/api/__init__.py
```diff
@@ -0,0 +1 @@
+default_app_config = 'metadeploy.api.apps.ApiConfig'
```
cdcarter: i believe specifying default_app_config is now unnecessary (if not actively discouraged). is this needed?

metadeploy/api/models.py
```diff
@@ -1,20 +1,101 @@
...
+
+    class Meta:
+        ordering = ('-created_at',)
+
+    def __str__(self):
+        return self.slug
+
+
+class ProductQuerySet(models.QuerySet):
+    def published(self):
```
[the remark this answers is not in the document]
jgerigmeyer: I'm going ahead and merging this... assuming this change can happen in a later PR, if at all.

PR #40, opened 2018-09-13
[better-allauth-logging] Log auth errors

wlonk:
zebra


wlonk:
@jgerigmeyer I think that I have a solution.

I'm using a custom adapter so that I can log errors and get them into sentry. But it'll still land on the template that Allauth uses for login errors. So I'm copying that template into the repo, and we can change it to be the same as the index template that renders the React app, possibly with some flag set to say "you had a login error". I am happy to talk about this more if you have questions, I doubt I'm explaining it well.

config/settings/base.py
```diff
@@ -314,10 +315,16 @@ def env(name, default=NoDefaultValue, type_=str):
 # Raven / Sentry
 SENTRY_DSN = env('SENTRY_DSN', default='')
 
-
+LOG_REQUEST_ID_HEADER = "HTTP_X_REQUEST_ID"
+LOG_REQUESTS = True
```
cdcarter: can we set 
```
GENERATE_REQUEST_ID_IF_NOT_IN_HEADER = True
REQUEST_ID_RESPONSE_HEADER = "X-Request-ID"
```
as well?

wlonk:
@jgerigmeyer Before we merge this, I'd love your input on the "don't show the raw allauth error page" approach I suggested, and perhaps to get that change in.

jgerigmeyer:
@wlonk That works for me. Should I go ahead and modify the `authentication_error.html` template?

wlonk:
@jgerigmeyer Please! I think that should do what we need, if you do. Do you want it to be just the same as `index.html`, or to include an error message variable or something?

wlonk:
@jgerigmeyer 👍

PR #42, opened 2018-09-14
Add python coverage data to coveralls.io

jgerigmeyer:
python


.circleci/config.yml
```diff
@@ -113,22 +113,21 @@ jobs:
...
             yarn test:check-coverage
       - store_artifacts:
           path: coverage
           destination: js-reports
       - store_test_results:
           path: coverage
+      - save_cache:
+          key: v1-metadeploy-jscoverage-{{ .Revision }}
+          paths:
+            - "~/metadeploy/coverage"
```
davisagli: Since the coverage results only need to be shared with subsequent jobs in the workflow and not with other builds, I think it may be more appropriate to use the `persist_to_workspace`/`attach_workspace` commands instead of `save_cache`/`restore_cache`. And then you don't have to worry about cache keys. This is what I set up for cumulusci a couple days ago: https://github.com/SFDO-Tooling/CumulusCI/blob/master/.circleci/config.yml

PR #50, opened 2018-09-19
[hotfix-precommit] 🐈 Use git-format-staged instead of lint-staged.

jgerigmeyer:
Fixes precommit with partially-staged files (e.g. using `git add -p`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "file-loader": "^2.0.0",
     "flow-bin": "^0.80.0",
     "flow-types": "^1.0.0",
+    "git-format-staged": "^2.1.0",
     "html-webpack-plugin": "^3.2.0",
     "http-proxy-middleware": "^0.19.0",
     "husky": "^0.14.3",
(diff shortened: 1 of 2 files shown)


jgerigmeyer:
This fixes issues with the previous tool (lint-staged) running the precommit linting hook on entire files, even if the files were only partially staged (e.g. `git add -p`). It seems to be a known bug (https://github.com/okonet/lint-staged/issues/62), but actively being worked on (https://github.com/okonet/lint-staged/pull/75).

This PR switches to a different tool (https://github.com/hallettj/git-format-staged) which fixes that particular issue, but also has much less useful output (e.g. does not tell you which file failed linting). I'm leaving this here, but it may be worth waiting to merge this for a bit to see if lint-staged actually fixes it soon -- if that's the case, I'd probably just close this.

jgerigmeyer:
@wlonk This problem should be fixed as of #89. Please test and close this PR if it's no longer an issue.

PR #52, opened 2018-09-20
[plan-detail] 🛢 Add plan-detail page.

jgerigmeyer:
https://trello.com/c/SIkvg5jw


wlonk:
**Before applying 5f22eec**, run `python manage.py migrate api 0023`, then run `python manage.py migrate` again after applying that diff.

src/js/components/plans/plan404.js
```diff
@@ -0,0 +1,33 @@
...
+const PlanNotFound = ({
+  product,
+  version,
+}: {
+  product: ProductType,
+  version: VersionType,
+}) => (
+  <FourOhFour>
+    <p>
+      We can&rsquo;t find the plan you&rsquo;re looking for. Try&nbsp;
```
wlonk: Can we not just include Unicode smart quotes in here? Or have a rendering layer that smartens simple quotes? I don't _super_ object to these entities, but they're not the height of legibility.

jgerigmeyer:
@wlonk Yeah, I think I'd lean that direction too... The trade-off is that then the markup is split over multiple components, with bits of JS intermixed. Early on @mirisuzanne and @stacyk specifically requested "one component per file" so that the markup is all together, but that does push somewhat toward larger components, or lots and lots and lots of very small files (which might make it even more difficult find things). 🤷‍♂️

wlonk:
OK, well that makes sense as a reason why we specifically would do it this way, then. Sorta Conway's Law-esque, makes sense to me.

src/js/components/404.js
```diff
@@ -15,7 +15,7 @@ const FourOhFour = ({ children }: { children?: React.Node }) => (
       <h1 className="slds-text-heading_large">Oh No!</h1>
       {children === undefined ? (
         <p>
-          That page cannot be found. Try the&nbsp;
+          That page cannot be found. Try the{' '}
```
wlonk: This is mildly gross, but I see why we have it. Eugh.

jgerigmeyer: Hey, you didn't like the entities... :)

src/sass/components/_steps.scss
```diff
@@ -0,0 +1,13 @@
+// Products
```
stacyk: Should the title here be steps? Super minor, but maybe confusing?

SondraE:
@jgerigmeyer Looking good. 

I like the use of the file box in place of the gift box for Managed Packages.

I'll keep playing with the layout of required/optional, but I'd say go ahead with this layout for now.

The name of the Plan really does get lost on the Plan Detail page. Seeing it here, I'd vote even more strongly for have the Plan name to appear in the page header in some way.

There's one thing I'd love to change, if possible. When I'm on the Product or Plan Detail page, and then I click "Product" nav or the logo to go back to the list of products, I am taken to the "community" tab even if I selected a product under the "salesforce" tab. Though it didn't take long to figure out that I had gone back to different tab, this was disorienting at first.

metadeploy/api/models.py
```diff
@@ -253,6 +254,7 @@ class Plan(SlugMixin, models.Model):
 
     title = models.CharField(max_length=128)
     version = models.ForeignKey(Version, on_delete=models.PROTECT)
+    preflight_message = models.CharField(blank=True, max_length=2048)
```
cdcarter: can has TextField and backlog an item to markdownify them?

cdcarter: no max_length needed on these messages

metadeploy/api/models.py
```diff
@@ -270,3 +272,41 @@ def natural_key(self):
...
+        ('data', _('Data')),
+    )
+
+    plan = models.ForeignKey(Plan, on_delete=models.PROTECT)
+    name = models.CharField(max_length=1024)
+    description = models.TextField()
+    is_required = models.BooleanField(default=True)
+    is_recommended = models.BooleanField(default=True)
+    kind = models.CharField(choices=Kind, default=Kind.metadata, max_length=64)
+    order_key = models.PositiveIntegerField(default=0)
```
cdcarter: any reason not to `order_with_respect_to` instead?

metadeploy/api/models.py
```diff
@@ -270,3 +272,41 @@ def natural_key(self):
...
+    def kind_icon(self):
+        if self.kind == self.Kind.metadata:
+            return 'package'
+        if self.kind == self.Kind.onetime:
+            return 'apex'
+        if self.kind == self.Kind.managed:
+            return 'archive'
+        if self.kind == self.Kind.data:
+            return 'paste'
+        return None
```
cdcarter: what if we just return self.kind in this case, and if it happens to be a valid utility icon, the frontend can do the needful, and if it doesn't, then we get what we deserve.

src/js/components/plans/detail.js
```diff
@@ -0,0 +1,137 @@
...
+        <BodyContainer>
+          <div
+            className="slds-text-longform
+              slds-p-around_medium
+              slds-size_1-of-1
+              slds-medium-size_1-of-2"
+          >
+            {plan.preflight_message ? (
+              <>
+                <h3 className="slds-text-heading_small">Preflight</h3>
```
cdcarter: this doesn't seem like its gonna be meaningful to the user

src/sass/components/_steps.scss
```diff
@@ -0,0 +1,13 @@
...
+// ========
+
+#plan-steps-table {
+  .slds-accordion__section {
+    padding-left: 0;
+    padding-right: 0;
+  }
+
+  .slds-accordion__summary-heading {
+    font-size: $font-size-medium;
```
stacyk: Originally I thought it might make sense to use Did we use `slds-text-heading_medium` as a class in the markup instead of `slds-accordion__summary-heading` but we would have to override the font-weight and line-height that comes with `slds-text-heading_medium` if we go that direction. I'd stick with this variable that you have here :)

src/js/components/plans/stepsTable.js
```diff
@@ -0,0 +1,175 @@
...
+      ) : null}
+      <span>{value}</span>
+    </DataTableCell>
+  );
+};
+KindDataCell.displayName = DataTableCell.displayName;
+
+const RequiredDataCell = (props: DataCellProps): React.Node => {
+  const required = props.item && props.item.is_required;
+  let classes = 'slds-align-middle slds-badge slds-m-horizontal_large';
```
stacyk: this spacing looks good with the large class.

metadeploy/api/models.py
wlonk: Not once they're `TextFields`, no. Will do.

metadeploy/api/models.py
wlonk: Because I missed it! I spent some time looking for a third-party package I remembered that did this, but it's in-Django now, I guess!

metadeploy/api/models.py
wlonk: I think that it's useful for the backend to give both the text to display, and the icon name to use, rather than relying on the frontend to guess it. It feels like a backend concern to me?

src/sass/components/_steps.scss
jgerigmeyer: Yeah, unfortunately we have very little control over the classes in the Accordion component, so I went this route instead as a matter of practicality.

src/js/components/plans/detail.js
cdcarter: the content itself should be enough IMO

metadeploy/api/models.py
cdcarter: understood re layout @jgerigmeyer, since we use the Choice() get_display_name stuff also, lets just leave it as is.

PR #53, opened 2018-09-21
[hotfix-prod-settings] Find overridden templates in production settings

wlonk:
Imagine a cute bunny.


wlonk:
@jgerigmeyer done.

PR #57, opened 2018-09-25
[https-always] Add settings to always use HTTPS

wlonk:
tree kangaroo

I've already set the env variables on staging; you won't need them on local dev.


davisagli:
Hmm, I think maybe it would be a better default to have this on by default except in the local and test settings?

PR #58, opened 2018-09-25
log rq job id when in a job

cdcarter:
cooper in a hoodie

Attach a logging filter (ala [using filters to impart contextual information](https://docs.python.org/2/howto/logging-cookbook.html#using-filters-to-impart-contextual-information), the same technique used by django_log_request_id) to the rq_console logging handler. It adds the current job_id to the JobRecord if we're in a job.

### DIFF
--- a/metadeploy/logfmt.py
+++ b/metadeploy/logfmt.py
@@ -1,8 +1,18 @@
-import datetime
 import numbers
+import logging
+import datetime
 
 from django.utils.log import ServerFormatter
+from rq import get_current_job
 
+NO_JOB_ID = 'no-job-id'
+
+class JobIDFilter(logging.Filter):
+    def filter(self, record):
+        if get_current_job():
+            record.job_id = get_current_job().id
+        else:
+            record.job_id = NO_JOB_ID
 
 class LogfmtFormatter(ServerFormatter):
     """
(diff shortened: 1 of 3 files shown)


cdcarter:
yea, test incoming.

If it weren't for django-log-request-id AND the logging cookbook having examples I wouldn't have gone the filter route, it sure is a little weird. But hey, its not run til after the loglevel decision is made which is nice for performance.

metadeploy/logfmt.py
```diff
@@ -82,7 +83,7 @@ def format(self, record):
         tag = self._get_tag(record)
         rest = self.format_line(getattr(record, 'context', {}))
         return ' '.join(filter(None, [
-            f'id={id_}',  # TODO: get-id method
+            f'id={id_}',
```
wlonk: Hah good catch thanks.

metadeploy/tests/logfmt.py
```diff
@@ -1,7 +1,55 @@
...
+    assert record.job_id == 123
+
+
+def test_job_id_filter__no_job(mocker):
+    get_id = mocker.patch('metadeploy.logfmt.get_current_job')
+    get_id.return_value = None
+    log_filter = JobIDFilter()
+    record = SimpleNamespace()
+    log_filter.filter(record)
+    assert record.job_id == 'no-job-id'
```
wlonk: This is non-blocking, but this is where I'd group a set of tests for various paths through a single call under a class (`TestJobIdFilter` or something).

jgerigmeyer:
Looks like the build is still failing here: `flake8`

PR #62, opened 2018-09-27
[retain-context] Retain user context while navigating (login, active-products-tab).

jgerigmeyer:
- https://trello.com/c/sHjvmvGN
- https://trello.com/c/CW6sYSj6


cdcarter:
@davisagli would you mind taking a review of this front-end only PR, and I'll do functional/acceptance testing when its in staging?

src/js/app/reducer.js
```diff
@@ -2,20 +2,24 @@
...
 export type AppState = {
   +user: User,
   +products: Products,
+  +settings: Settings,
 };
 
 const reducer = combineReducers({
   user: userReducer,
   products: productsReducer,
+  settings: settingsReducer,
```
wlonk: I have no great objection to this mapping here, but I've also seen just naming the reducers `settings`, `user`, etc, importing that, and just using the key-is-the-same-as-variable-name syntax.

src/js/components/header/customDomainModal.js
```diff
@@ -28,7 +30,12 @@ class CustomDomainModal extends React.Component<Props, { url: string }> {
       return;
     }
     const baseUrl = window.api_urls.salesforce_custom_login();
-    window.location.assign(`${baseUrl}?custom_domain=${val}`);
+    window.location.assign(
+      addUrlParams(baseUrl, {
+        custom_domain: val,
+        next: window.location.pathname,
+      }),
```
wlonk: Ooh, I like this pattern.

jgerigmeyer:
There would be a number of options for where to store this data (the recently-active tab) -- in memory, a URL hash, or localStorage/IndexedDB. I chose the cache (IndexedDB) for a few reasons:

1. We want it to persist through an oauth-login-redirect flow, which makes the in-memory option less appealing.
2. We want it to persist when you click the "Products" link in the header to go back to the list page, which makes the URL hash option less appealing.

This means the cached default-active-tab persists indefinitely, until the cache is cleared on logout. We could change this behavior -- or add a URL hash in addition -- but it seemed like a reasonable place to start.

src/js/app/reducer.js
davisagli: +1. The term reducer is kinda a pet peeve of mine (what are we making smaller, exactly?)

src/js/components/products/list.js
```diff
@@ -7,18 +7,35 @@ import TabsPanel from '@salesforce/design-system-react/components/tabs/panel';
...
+    productsByCategory: ProductsMapType,
+    productCategories: Array<string>,
+    activeProductsTab: ActiveProductsTabType,
+    doSaveActiveTab: typeof saveActiveTab,
+  },
+  { activeProductsTab: ActiveProductsTabType },
+> {
+  constructor(props) {
+    super(props);
+    this.state = { activeProductsTab: props.activeProductsTab };
```
davisagli: I'm new to redux so maybe I'm missing something, but I don't see anything that updates the component state after it's constructed?

src/js/components/products/list.js
```diff
@@ -62,7 +87,18 @@ class ProductsList extends React.Component<{
...
         }
-        contents = <Tabs variant="scoped">{tabs}</Tabs>;
+        const savedTabIndex = this.props.productCategories.indexOf(
+          this.state.activeProductsTab,
+        );
+        contents = (
+          <Tabs
+            variant="scoped"
+            onSelect={this.handleSelect}
+            defaultSelectedIndex={savedTabIndex === -1 ? 0 : savedTabIndex}
```
davisagli: Since you're only setting the Tabs component's *default* selected index I think you are relying on all mutations to the selected index to happen via interaction with the tabs DOM (which triggers onSelect which dispatches the action). If something else in the app were to change the activeProductsTab, it doesn't look to me like that the Tabs component would actually update to reflect that. I think what you want to do is set `savedTabIndex` based on this.props rather than this.state, and pass selectedIndex to Tabs instead of defaultSelectedIndex, so that the Tabs state is controlled from outside instead of tracked internally. Correct me if I'm wrong; I'm newish to this too.

src/js/utils/caching.js
```diff
@@ -8,7 +8,7 @@ export const cache = getConfiguredCache({
 });
 
 const actionsToPersist = {
-  // ...
+  PRODUCTS_TAB_ACTIVE: ['settings'],
 };
```
davisagli: Wow, this is kinda magical (in a good way, I think)! How is browser support for IndexedDB?

src/js/components/products/list.js
jgerigmeyer: So, I did it this way specifically so that the saved setting would _only_ impact initial render, and then the user/DOM would take over. I guess I couldn't think of a use-case for wanting the tab to change programmatically (i.e. making it a controlled component). For example, if you have the app open in two tabs, your proposal would mean that they would always stay in sync -- is that a better UX?

I don't feel strongly about this though, and I guess I could see arguments both ways. If you prefer the controlled component route, I could be convinced.

src/js/utils/caching.js
jgerigmeyer: Agreed! It's using https://github.com/HenrikJoreteg/redux-persist-middleware and https://github.com/HenrikJoreteg/money-clip...

Browser support is pretty good (https://caniuse.com/#feat=indexeddb), but there are polyfills if it becomes an issue. (It also fails silently if the cache doesn't work.)

jgerigmeyer:
@davisagli Agreed re: the URL hash -- that's exactly why it would be useful. Then again, that's a distinct use-case, and not the one this is trying to address. It would be simple to add down the road if we want that as well (and the URL would just take precedence over the cached value).

src/js/components/products/list.js
davisagli: Oh huh, I hadn't thought about the two-browser-tabs thing (partly because I was just starting to understand how this stuff gets persisted). But let's see...currently if I switch to Component Tab B in browser tab 1, then switch to Component Tab A in browser tab 2, then in browser tab 1 hit the back button, it will take me back to Component Tab A rather than where I used to be in browser tab 1, right? This is a bit contrived but it makes me think it a bit weird that this gets set browser-wide rather than per-browser-tab. Maybe it should go in `window.sessionStorage` instead of indexeddb?

Also, couldn't you just rely on using defaultSelectedIndex to avoid controlling the component, and pass it the setting from this.props rather than this.state?

jgerigmeyer: 1. Hm, that's a good point. I was considering in-memory and localStorage/indexedDB, but didn't seriously consider sessionStorage... I guess it comes down to how widely/long we want this to persist, but thinking through the use-cases (browser back button, login-redirect, clicking nav link to go back) I'm liking that proposal.

2. Yes, as long as `defaultSelectedIndex` changing doesn't impact the rendered markup. But it doesn't, so... yes.

jgerigmeyer:
@davisagli I'd appreciate your review of this latest iteration.

src/js/index.js
```diff
@@ -63,7 +63,22 @@ const App = () => (
...
+                  } catch (e) {
+                    // swallow error
+                  }
+                  return (
+                    <ProductsList
+                      {...props}
+                      activeProductsTab={activeProductsTab}
+                    />
+                  );
+                }}
```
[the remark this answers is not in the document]
jgerigmeyer: Same as before... this is called on browser back/forward navigation, so the cached value is passed in as a prop every time it is mounted.

Would you prefer that it calculates internally and stores in the component state on `constructor` or `componentWillMount`? I went back and forth, but since the value never needs to change, a `prop` passed in seemed potentially More Righter. 🤷‍♂️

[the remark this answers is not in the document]
davisagli: What ties selecting a tab to creating a browser history entry? And when you hit back how does this get the previous value from sessionStorage instead of the one that was just saved when you selected a new tab? This is no longer tied to the redux state unless I'm missing something.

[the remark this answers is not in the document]
jgerigmeyer: Oh, maybe I misunderstood your question. You're correct that this is no longer tied to Redux state. Changing the tab does not create a new browser history entry (I wasn't thinking that was really part of this feature request). If you navigate away from this page, you can use browser-back and it correctly remembers the selected tab -- same with refreshing the page.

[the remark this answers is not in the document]
davisagli: Oh yeah, got it! That's good enough for now.

I do like the idea of keeping this internal to the ProductsList component state, since it doesn't seem like a concept that the router should need to know about. But I'll leave that up to you.

PR #63, opened 2018-09-27
[start-job] 🛢💥 Attempt to make Job endpoint align with needs

wlonk:
hippo

This is a tentative first start, I think a lot of pieces are up for reconsideration.

**BIG IMPORTANT NOTICE!**

This PR declares migration bankruptcy. You will have to run:

```
dropdb metadeploy
createdb metadeploy
python manage.py migrate
python manage.py populate_sample_products
python manage.py populate_social_apps --prod-id […] --test-id  […] --cust-id  […] --prod-secret  […] --test-secret  […] --cust-secret  […]
```


metadeploy/api/jobs.py
```diff
@@ -130,17 +130,18 @@ def run_flow(user, repo_url, flow_name):
...
-            proj_config,
-            flow_config,
-            proj_keychain.get_org(current_org),
-            options={},
-            skip=[],
-            name=flow_name,
-        )
-        flowinstance()
+        # Make and run the flows:
+        for flow_name in flow_names:
```
wlonk: Is this a reasonable approach here?

metadeploy/api/models.py
```diff
@@ -129,6 +130,7 @@ class Product(SlugMixin, models.Model):
         max_length=32,
     )
     slds_icon_name = models.CharField(max_length=64, blank=True)
+    repo_url = models.URLField(blank=True)
```
wlonk: Should this be on the `Product` or the `Version`?

metadeploy/api/models.py
```diff
@@ -289,6 +294,7 @@ class Step(models.Model):
     is_recommended = models.BooleanField(default=True)
     kind = models.CharField(choices=Kind, default=Kind.metadata, max_length=64)
     order_key = models.PositiveIntegerField(default=0)
+    flow_name = models.CharField(max_length=64, blank=True)
```
wlonk: This is `blank=True` just to aid the migration; I'm not sure what a coherent data migration to populate it would be though.

metadeploy/api/serializers.py
```diff
@@ -93,16 +93,58 @@ class Meta:
...
+        queryset=Step.objects.all(),
+        many=True,
+    )
+
+    def create(self, validated_data):
+        repo_url = validated_data.pop('plan').version.product.repo_url
+        validated_data['repo_url'] = repo_url
+        flow_names = [step.flow_name for step in validated_data.pop('steps')]
+        validated_data['flow_names'] = flow_names
+        return super().create(validated_data)
```
wlonk: This serializer shows, like, three different approaches to similar problems. I wanna come back to it and unify the approaches, and figure out what information it should surface when you later view the `Job` in the API.

metadeploy/api/models.py
cdcarter: yes.

cdcarter: ;)

cdcarter: version specifies the tag, product specifies the URL? right now we're just checking out master, right?

metadeploy/api/serializers.py
```diff
@@ -93,16 +93,58 @@ class Meta:
...
 
+class InstanceUrl:
+    def set_context(self, serializer_field):
+        # TODO: This is brittle, has room for runtime errors.
+        self.instance_url = (
+            serializer_field
+            .context['request']
+            .user
+            .socialaccount_set.first()
+            .extra_data['instance_url']
```
cdcarter: this feels like model logic

metadeploy/api/serializers.py
```diff
@@ -93,16 +93,58 @@ class Meta:
...
+    )
+    steps = serializers.PrimaryKeyRelatedField(
+        queryset=Step.objects.all(),
+        many=True,
+    )
+
+    def create(self, validated_data):
+        repo_url = validated_data.pop('plan').version.product.repo_url
+        validated_data['repo_url'] = repo_url
+        flow_names = [step.flow_name for step in validated_data.pop('steps')]
```
cdcarter: we're presumably going to want to reference more information from the step in the UI, so i have a hunch that this is going to end up using step keys instead of flow names

metadeploy/api/models.py
wlonk: Oh ho, I see. So, I'm content leaving `repo_url` on `Product`, and we'll add a `tag` field to version or use the label or something, to build the whole URL.

metadeploy/api/jobs.py
davisagli: I don't see why not, as long as this method gets renamed to `run_flows`

metadeploy/api/models.py
wlonk: So, I've moved Jobs to have the user, plan, steps, and to decompose the necessary information off of that. Related to this in particular, I've got the `repo_url` on the `Product` now, and the `commit_ish` on the `Version`. Let me know what you think.

metadeploy/api/jobs.py
```diff
@@ -90,7 +84,10 @@ def run_flow(user, repo_url, flow_names):
         stack.enter_context(prepend_python_path(os.path.abspath(tmpdirname)))
 
         # Let's clone the repo locally:
-        git.Repo.clone_from(repo_url, tmpdirname)
+        # Split commit-ish off:
+        repo = git.Repo.clone_from(repo_url, tmpdirname)
+        # Uncertain about this line; is there a better way?
+        getattr(repo.heads, commit_ish).checkout()
```
cdcarter: on other projects it looks like we've used the library "github3.py" since all of our products are hosted on github (or conceivably a GHE instance). @jlantz may have opinions about if we have any reason to support more than GitHub...

but if we use the github api, we can get the archive zip for a specific commit/ref REALLY easily.

I think that doing a deep clone of Cumulus/HEDA every time is not a great idea, they have such complicated git histories. I might even go so far as to say: we should cache the latest versions of our products in somewhere closer to the worker (backlog, possibly beyond v1). But, unless we have a reason not to, lets use githubs infra to extract the files we want, not our own.

metadeploy/api/jobs.py
```diff
@@ -144,16 +141,16 @@ def run_flow(user, repo_url, flow_names):
...
-        j.job_id = run_flow_job.delay(
+        j.job_id = run_flows_job.delay(
             j.user,
-            j.repo_url,
-            j.flow_names,
+            j.plan,
+            j.steps,
         ).id
         j.enqueued_at = timezone.now()
         j.save()
```
cdcarter: all of our RQ jobs need to close their connection at the end of the work. the rqworker management command is nice enough to close the db it holds conn before fork()ing the workhorse but it doesn't clean up when its done. we should backlog an item for a custom worker class that closes its conn.

metadeploy/api/jobs.py
wlonk: I'll look into `github3.py`, the library I'm currently using (GitPython) hasn't won me over.

Let's make another story about caching.

metadeploy/api/models.py
```diff
@@ -205,6 +196,7 @@ class Version(models.Model):
     description = models.TextField()
     created_at = models.DateTimeField(auto_now_add=True)
     is_production = models.BooleanField(default=True)
+    commit_ish = models.CharField(max_length=256, default='master')
```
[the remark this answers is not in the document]
wlonk: Per the git docs, a branch/tag/commit is usually called a "commit-ish", so I went with their terminology.

metadeploy/api/jobs.py
wlonk: I'd be interested to hear more about that, in that backlog item.

metadeploy/api/jobs.py
cdcarter: Yea, GitPython isn't particularly gorgeous. After some conversation in our team chat this morning, I'm not concerned about supporting anything other than github, so using github3.py (or any other github api wrapper) to get the zipball would be our preferred solution. particularly so that we can retrieve PRIVATE repos.

wlonk: I'll work on getting that library into this PR.

Let's talk (on Slack?) about private repos, because I think there's more there.

jgerigmeyer:
@wlonk Because of the changes to migrations, it will be painful to switch on/off this branch. Would it be possible to get this to a merge-able state soon, then continue more jobs-specific work on another PR?

jgerigmeyer:
I guess alternatively we could just cherry-pick and merge the migration-specific changes...

metadeploy/api/models.py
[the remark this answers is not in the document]
cdcarter: fair, I'm just not sure if it's ever going to be a commit, so although standard, its also counterintuitive to our use. even if we ever did want to pin to a specific commit, we'd do it with a tag... 

let's leave it called this, but for me in three years can you throw some helptext reminding that this is frequently a branch/tag. just so nobody naively sees commit and hopes that it'll usually start with the hash they're looking for.

metadeploy/api/jobs.py
```diff
@@ -130,29 +127,30 @@ def run_flow(user, repo_url, flow_name):
...
     for j in Job.objects.filter(enqueued_at=None):
-        j.job_id = run_flow_job.delay(
+        j.job_id = run_flows_job.delay(
             j.user,
-            j.repo_url,
-            j.flow_name,
+            j.plan,
+            j.steps,
         ).id
         j.enqueued_at = timezone.now()
```
cdcarter: consider:
```
for j in Job.objects.filter(enqueued_at=None):
    rq_job = run_flows_job.delay(j.user, j.plan, j.steps)
    j.job_id = rq_job.id
    j.enqueued_at = rq_job.enqueued_at
    j.save()
```

cdcarter: I started writing a long comment here and then it got lost. I'm not sure about the transactionality of this, and I think, ideally, we want to do something like COMPLETE AND TOTAL PSEUDOCODE THAT CAN'T POSSIBLY WORK:
```
job_set = Job.objects.filter(enqueued_at=None).select_for_update
jobs = [ {job_model_id:  rq_job.job_id} for(job_model_id, rq_job) in map(lambda j: {j.id: run_flows_job(**j.stuff_to_enqueue)} ]
Job.objects.filter(id__in=job_set).update(**jobs)
```

wlonk: So, my difficulty with django-rq is that the docs are a bit sparse; is there somewhere I can read about `.delay` and what it returns, and what that object might have, so that I would know this? In the readme at https://github.com/rq/django-rq, there's no elaboration on `.delay`.

wlonk:
@cdcarter I'd like to merge this, then work on the `github3.py` stuff, if that's OK with you.

wlonk:
(Also, I am OK with merging this and then doing the related frontend things in a different PR if @jgerigmeyer is!)

PR #71, opened 2018-09-30
[newrelic-rq] Add New Relic monitoring for django-rq worker/scheduler.

jgerigmeyer:
Also some minor fixes/improvements to CircleCI config.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 web: newrelic-admin run-program gunicorn metadeploy.wsgi:application --preload
-worker: python manage.py rqworker default short
-scheduler: python manage.py rqscheduler default short
+worker: newrelic-admin run-python manage.py rqworker default short
+scheduler: newrelic-admin run-python manage.py rqscheduler default short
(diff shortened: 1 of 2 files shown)


Procfile
```diff
@@ -1,3 +1,3 @@
 web: newrelic-admin run-program gunicorn metadeploy.wsgi:application --preload
-worker: python manage.py rqworker default short
-scheduler: python manage.py rqscheduler default short
+worker: newrelic-admin run-python manage.py rqworker default short
+scheduler: newrelic-admin run-python manage.py rqscheduler default short
```
jgerigmeyer: These commands seem to work, but I'm not sure how to test them functionally.

cdcarter: if you load `/admin/django-rq/` while the worker is up it should appear for each queue in the workers column.

sadly best way to check for scheduler atm is checking redis for the key, or try to bring up another scheduler instance and let it fail.

i'd love to get a PR back into django-rq someday to add scheduler info to the monitoring page but it hasn't been a priority...

PR #76, opened 2018-10-02
[github3] Move to use Github3.py

wlonk:
red panda

### DIFF
--- a/metadeploy/api/jobs.py
+++ b/metadeploy/api/jobs.py
@@ -65,6 +65,18 @@ def prepend_python_path(path):
         sys.path = prev_path
 
 
+def is_safe_path(path):
+    return not os.path.isabs(path) and not path.startswith('..')
+
+
+def zip_file_is_safe(zip_file):
+    return all(
+        is_safe_path(info.filename)
+        for info
+        in zip_file.infolist()
+    )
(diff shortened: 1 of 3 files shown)


metadeploy/api/jobs.py
```diff
@@ -83,10 +91,14 @@ def run_flows(user, plan, steps):
...
-        # Split commit-ish off:
-        repo = git.Repo.clone_from(repo_url, tmpdirname)
-        # Uncertain about this line; is there a better way?
-        getattr(repo.heads, commit_ish).checkout()
+        gh = github3.login(token=settings.GITHUB_TOKEN)
+        user, repo_name = extract_user_and_repo(repo_url)
+        repo = gh.repository(user, repo_name)
+        zip_file_name = 'archive.zip'
+        repo.archive('zipball', path=zip_file_name, ref=commit_ish)
+        # TODO: This is not actually safe! It's safe by accident of
```
cdcarter: what is the TODO here? walk the archive before extracting it and make sure that everything is a relative path and not a banned mime type?

wlonk: Yes, thank you for catching that, I forgot I had left that TODO.

PR #77, opened 2018-10-03
Add custom RQ worker that closes DB connections

wlonk:
The tests are mock-heavy, but I think accurate.

elephant


cdcarter:
Yes please!

PR #79, opened 2018-10-05
[preflight-backend] 🛢 Run preflight checks.

wlonk:
snow leopard


wlonk:
Let's hand-test this against the preflight job that @cdcarter wrote before merging it:

> okay, i ONLY report `ok`s for now, but try out this branch https://github.com/SFDO-Tooling/CumulusCI-Test/tree/feature/preflight
https://github.com/SFDO-Tooling/CumulusCI-Test/blob/feature/preflight/cumulusci.yml#L44
> the flow is named `static_preflight`
> i'll add one that fails a step too

metadeploy/api/flows.py
```diff
@@ -0,0 +1,39 @@
...
+class PreflightFlow(flows.BaseFlow):
+    def __init__(self, *args, preflight_result=None, **kwargs):
+        self.preflight_result = preflight_result
+        return super().__init__(*args, **kwargs)
+
+    # def pre_flow(self, *args, **kwargs):
+    #     pass
+
+    def post_flow(self):
+        results = {
```
cdcarter: not 100% sure this destructuring is correct for the flow sample I came up with. doesn't block merge, because we'll need to come together on the format, I just guessed a bit.

metadeploy/api/jobs.py
```diff
@@ -80,22 +86,25 @@ def zip_file_is_safe(zip_file):
...
-    # post_subflow
-    #
-    # Can we do anything meaningful with a return value from a @job,
-    # too?
+    # Can we do anything meaningful with a return value from a @job?
+
+    # This is a little stupid, but it lets us mock out BasicFlow in the
+    # tests better than if we put this straight in the kwargs:
+    if flow_class is None:
+        flow_class = BasicFlow
```
cdcarter: i think you can just set this as a default, since its not mutable.

metadeploy/api/jobs.py
```diff
@@ -80,22 +86,25 @@ def zip_file_is_safe(zip_file):
...
 
     token, token_secret = user.token
     instance_url = user.instance_url
     repo_url = plan.version.product.repo_url
     commit_ish = plan.version.commit_ish
-    flow_names = [step.flow_name for step in steps]
+
+    if preflight_result:
```
cdcarter: i'm a bit confused abut what ... is happening here. what's preflight_result? is it the dict of results?

metadeploy/api/jobs.py
```diff
@@ -158,31 +167,42 @@ def run_flows(user, plan, steps):
...
+
 def enqueuer():
     logger.debug('Enqueuer live', extra={'tag': 'jobs.enqueuer'})
     for j in Job.objects.filter(enqueued_at=None):
-        rq_job = run_flows_job.delay(
-            j.user,
-            j.plan,
-            list(j.steps.all()),
-        )
+        skip_tasks = calculate_skips_for_plan(j.plan, j.steps.all())
```
cdcarter: i find it weird that this happens at the point of the enqueuer, not when the Job is created and not at the worker. I don't really like the enqueuer having more purpose than needed.

metadeploy/api/jobs.py
```diff
@@ -199,3 +219,23 @@ def expire_user_tokens():
 
 
 expire_user_tokens_job = job(expire_user_tokens)
+
+
+def preflight(user, plan):
+    preflight_result = PreflightResult.objects.create(
+        user=user,
+        plan=plan,
+        organization_url=user.instance_url,
```
cdcarter: is this to identify the specific org? if so, we should use the org_id (starts with 00D, i assume its somewhere in the token or extra infor...) - instance_url is non-unique in a lot of cases.

metadeploy/api/jobs.py
wlonk: This is for test support, alas. See what I wrote about it here: https://transneptune.net/2018/10/16/fun-times-with-unittest-mock/

wlonk: Actually in the commit that's about to land, there's a bunch of stuff I don't like that exists to help isolate IO away from tests, to maintain test coverage. I'd love to not have all this testing-mishegas in the code, though, and very much want to revisit this to try to restructure it to take that need out.

metadeploy/api/jobs.py
wlonk: It's a `PreflightResult` model, which, if present, means we're running this as a preflight job. I'm not committed to this, but I wanted to reuse as much of the Cumulus-running logic as I could; I'd like to eventually restructure it to be more decomposed and thus composable.

metadeploy/api/jobs.py
wlonk: I agree.

metadeploy/api/jobs.py
wlonk: I don't actually recall; it was called `organization_url`, which implied to me it was a uniquely-identifying URL for the associated org. I don't think we use this anywhere yet, though, so I'm not sure what it has to be.

wlonk:
OK, I have a lot of dissatisfactions with the most recent commit, but it helped the end to end hand-testing.

Part of what I really don't like is that this requires so much IO that I have to mock out, but I definitely don't want the tests touching actual Salesforce orgs, so I don't know that we can eliminate the mocking.

If we can reduce the mocking significantly, I think we can improve this code by removing test-motivated mishegas, while also maybe having tests that would actually catch the broken things that I encountered when hand-testing.

wlonk:
OK, I hand tested it as follows:

Run `yarn serve`. Freshly log in against a scratch org to have a valid token. Run the following in a Django shell:

```
from django.contrib.auth import get_user_model
from metadeploy.api.jobs import preflight_job
from metadeploy.api.models import Plan
User = get_user_model()
user = User.objects.all().last()
plan = Plan.objects.filter(title__icontains='prefli').first()  # This is a Plan I made that runs the static_preflight flow from Christian's links above. It uses a Product and Version that use that repo, and that branch.
preflight_job.delay(user, plan)
```

See job finish successfully.

jgerigmeyer:
Merging; further cleanup can happen in `preflight-results` branch.

PR #81, opened 2018-10-08
[require-login] 📦 🐈 Require valid token before running preflight.

jgerigmeyer:


requirements/base.txt
```diff
@@ -2,8 +2,10 @@ Django==2.1.2
 psycopg2-binary==2.7.5
 dj-database-url==0.5.0
 whitenoise==4.1
+channels==2.1.3
+channels_redis==2.3.0
+service_identity==17.0.0
 django-allauth==0.38.0
-django-extensions==2.1.3
```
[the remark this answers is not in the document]
wlonk: I think if you want `shell_plus` locally you can install this package yourself, though, right? Like, we don't have ipython in here, but if it's there, the `shell` command will pick it up, for instance.

I generally think that local-dev improvements don't live in requirements, because everyone likes different things.

jgerigmeyer:
@stacyk We're still working on adding realtime updates via websockets, but I think this PR is ready for design and Sass code review.

src/sass/components/_steps.scss
```diff
@@ -10,4 +10,13 @@
...
     font-size: $font-size-medium;
   }
+
+  .step-column-tooltip {
+    text-transform: none;
+    letter-spacing: 0;
+  }
+
+  .step-name {
+    margin-left: 1.375rem;
```
stacyk: I'd add a comment to why this value if it is a one off. Otherwise it is a mysterious magic number (likely required to line something up if I had to guess).

stacyk: yep, makes sense to me.

CONTRIBUTING.rst
```diff
@@ -108,16 +108,18 @@ To run the local development server::
...
-The WebSocket server (to live-reload on file changes) listens on
-``wss://localhost:5000``. In Firefox, this requires first visiting
-`<https://localhost:5000/>`_ to add another local certificate exception.
+To run the background worker tasks, ensure you have Redis running
+(possibly installing it via ``brew install redis``), then run each of
+the following in separate terminals, with the virtualenvironment
+active::
+
+   python manage.py rqworker default short
+   python manage.py rqscheduler default short
```
wlonk: @jgerigmeyer do you think we can/should roll these into `yarn serve`?

src/js/index.js
```diff
@@ -125,6 +126,11 @@ cache
         if (user) {
           // Login
           appStore.dispatch(login(user));
+          // Connect to WebSocket server
+          window.socket = createSocket({
+            url: 'ws://localhost:8080/ws/notifications/',
```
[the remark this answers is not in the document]
wlonk: This can pull from `window.api_urls.ws_notifications()` now.

metadeploy/api/jobs.py
```diff
@@ -189,3 +196,20 @@ def enqueuer():
...
+def expire_user_tokens():
+    token_lifetime_ago = timezone.now() - timedelta(
+        minutes=settings.TOKEN_LIFETIME_MINUTES,
+    )
+    tokens_to_expire = SocialAccount.objects.filter(
+        last_login__lte=token_lifetime_ago,
+    )
+    for sa in tokens_to_expire:
+        sa.socialtoken_set.all().delete()
+        async_to_sync(user_token_expired)(sa.user)
```
jgerigmeyer: @wlonk This seems to trigger a `user_token_expired` message every minute for every user who doesn't currently have a valid token. Could we only trigger it once, when the token goes from valid to invalid?

wlonk: Ew, gross, who would write a job that just spews that message.

I'll fix it. While I'm there, I'm gonna push this model logic into the models.

jgerigmeyer:
@wlonk Okay, as far as I'm aware this is good to go. Still interested in review from @cdcarter and/or @davisagli when they get a chance, but I'd say let's go ahead and merge/deploy -- I anticipate there may be unexpected things that crop up in deployment or on staging, and I'd like to test it out.

wlonk:
See also https://github.com/SFDO-Tooling/metadeploy/pull/83 and https://github.com/SFDO-Tooling/metadeploy/pull/84 where some hotfixes went.

src/js/components/plans/stepsTable.js
```diff
@@ -42,7 +44,7 @@ class NameDataCell extends React.Component<
             </AccordionPanel>
           </Accordion>
         ) : (
-          name
+          <span className="step-name-no-icon">{name}</span>
```
mirisuzanne: or `slds-p-left_large` (could also be `*-m-*`)?

src/js/components/plans/stepsTable.js
```diff
@@ -109,9 +115,7 @@ const InstallDataColumnLabel = (): React.Node => (
     <Tooltip
       align="top right"
       content={
-        <span style={{ textTransform: 'none', letterSpacing: 0 }}>
-          Select steps to install.
-        </span>
+        <span className="step-column-tooltip">Select steps to install.</span>
```
mirisuzanne: If `slds-text-title_caps` could be applied to the "install" span, it wouldn't have to be overridden.

src/sass/components/_steps.scss
mirisuzanne: It looks to me like this can be handled with slds-large padding or margin.

src/sass/components/_steps.scss
```diff
@@ -10,4 +10,15 @@
   .slds-accordion__summary-heading {
     font-size: $font-size-medium;
```
mirisuzanne: Wrong PR, I know – but I bet slds has a class for this?

src/js/components/plans/detail.js
```diff
@@ -78,27 +83,40 @@ const PlanDetail = ({
...
+                disabled={!plan.steps.length}
+              />
+            ) : (
+              <Login
+                id="plan-detail-login"
+                buttonClassName="slds-size_full
+                  slds-p-vertical_xx-small"
+                buttonVariant="brand"
+                triggerClassName="slds-size_full"
+                label="Start Pre-Install Validation"
```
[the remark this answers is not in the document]
jgerigmeyer: We actually started there, but it seemed pretty long and maybe not needed? I'm fine either way, if you think it's clearer.

src/js/components/plans/stepsTable.js
jgerigmeyer: Currently `.slds-text-title_caps` is being applied to the wrapping `th`, and we don't control that -- it's internal to design-system-react. That said, this is changing with the upcoming removal of title-caps from SLDS entirely (see https://github.com/salesforce/design-system-react/issues/1601).

src/sass/components/_steps.scss
jgerigmeyer: That's true (`.slds-text-body_regular`), but we don't have access to the right element to add a class.

src/js/components/plans/detail.js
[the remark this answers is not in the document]
mirisuzanne: You could drop "validation" from the long one. I think it's useful to find something that makes it clear what the button does. But I can be overruled on that.

src/js/components/plans/stepsTable.js
mirisuzanne: oh, maybe `-large` was just working on my screen by luck. Ok.

jgerigmeyer: Well it's darn close (`1.5rem` vs `1.375rem`). 🤷‍♂️

PR #82, opened 2018-10-08
[downgrade-github3] 📦 Downgrade github3.py

wlonk:
Until cumulusci can use the most recent version of it.

jerboa


wlonk:
Looks like our use of github3.py doesn't require `>=1.0.0`, so this is fine.

PR #97, opened 2018-10-29
[job-status] 🛢 Track completed steps on Job model

wlonk:
https://trello.com/c/nrONKfRn

And push when they change on the model.

capybara


wlonk:
Outstanding questions:

 - [ ] How does this help us resume a partial install?
 - [ ] How does the frontend get the job ID back so it can view progress, given that the job is sent through an enqueuer?

wlonk:
Answers:

 - "Resume" is just "Trigger the job again, with a skip list of "all the steps you've already completed". This will require some backend work to ensure that if there's a (recent?) job that already shows some steps completed, then those steps are always skippable on new jobs.
 - When you POST to make a new Job, you get back the Job's `id`. You can then query that endpoint (or listen to the websocket) to get updates to the `completed_steps` attribute.

wlonk:
OK, 2. is done in this branch.

I think 1. will require actually enforcing required steps, and then excepting that based on the previously run jobs.

wlonk:
I'd love to merge this before worrying about the frontend piece, as I've discovered that there's work in #93 that relies on it, and sending additional information to the frontend shouldn't break anything.

PR #101, opened 2018-11-01
[logged-in-as] Expose Org Name and Org Type on User serializer

wlonk:
zebra

### DIFF
--- a/metadeploy/multisalesforce/views.py
+++ b/metadeploy/multisalesforce/views.py
@@ -17,15 +17,27 @@ from .provider import (
 
 class SaveInstanceUrlMixin:
     def complete_login(self, request, app, token, **kwargs):
-        resp = requests.get(self.userinfo_url, params={'oauth_token': token})
+        resp = requests.get(self.userinfo_url, params={"oauth_token": token})
         resp.raise_for_status()
         extra_data = resp.json()
-        instance_url = kwargs.get('response', {}).get('instance_url', None)
+        instance_url = kwargs.get("response", {}).get("instance_url", None)
         ret = self.get_provider().sociallogin_from_response(
             request,
             extra_data,
         )
-        ret.account.extra_data['instance_url'] = instance_url
+        ret.account.extra_data["instance_url"] = instance_url
+        org_url = (
+            extra_data["urls"]["sobjects"] + "Organization/{org_id}"
+        ).format(
+            version="44.0",
+            org_id=extra_data["organization_id"],
+        )
+        headers = {
+            "Authorization": "Bearer {}".format(token),
+        }
+        resp = requests.get(org_url, headers=headers)
+        resp.raise_for_status()
+        ret.account.extra_data["organization_details"] = resp.json()
         return ret
 
 
--- a/metadeploy/api/models.py
+++ b/metadeploy/api/models.py
@@ -40,6 +40,24 @@ class User(AbstractUser):
     objects = UserQuerySet.as_manager()
 
     @property
+    def org_name(self):
+        if self.social_account:
+            return self.social_account.extra_data.get(
+                'organization_details',
+                {},
+            ).get('Name')
+        return None
+
+    @property
+    def org_type(self):
+        if self.social_account:
+            return self.social_account.extra_data.get(
+                'organization_details',
+                {},
+            ).get('OrganizationType')
+        return None
+
+    @property
     def instance_url(self):
         if self.social_account:
             return self.social_account.extra_data.get('instance_url', None)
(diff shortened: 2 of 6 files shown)


jgerigmeyer:
I think the JS/front-end part of this is complete, so just waiting on whatever cleanup and permissions error-handling @wlonk intends to do.

metadeploy/multisalesforce/views.py
```diff
@@ -16,16 +16,49 @@
...
+            user_id=extra_data["user_id"],
+        )
+        resp = requests.get(user_url, headers=headers)
+        resp.raise_for_status()
+        profile_id = resp.json()["ProfileId"]
+        profile_url = (sobjects_url + "Profile/{profile_id}").format(
+            profile_id=profile_id,
+        )
+        resp = requests.get(profile_url, headers=headers)
+        resp.raise_for_status()
```
cdcarter: yeehaw, instead of hitting the sobject endpoint for the profile...

org_info_url = (extra_data["urls"]["rest"] + "connect/organization").format(version="44.0")

You'll get an "Organization" response https://developer.salesforce.com/docs/atlas.en-us.chatterapi.meta/chatterapi/connect_responses_organization.htm, which has a "userSettings" object which has a key called... "canModify​AllData" which is the aggregate permission!

That org response also includes the org name, which should be sufficient for the multisalesforce plugin in general.

metadeploy/multisalesforce/views.py
```diff
@@ -16,16 +16,49 @@
...
+        # TODO: Don"t use assert outside of tests.
+        assert resp.json()["PermissionsModifyAllData"]
+        org_url = (sobjects_url + "Organization/{org_id}").format(
+            org_id=extra_data["organization_id"],
+        )
+        resp = requests.get(org_url, headers=headers)
+        resp.raise_for_status()
+        return resp.json()
+
     def complete_login(self, request, app, token, **kwargs):
```
cdcarter: it should be possible to put some or all of these calls into one composite or batch http request: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite.htm

metadeploy/multisalesforce/views.py
wlonk: Cool! Thanks for digging into that. Yay for actual correct permissions.

metadeploy/multisalesforce/views.py
wlonk: Thanks, I'll read up on that.

wlonk: Composite doesn't support the connect endpoints, and batch complained that GET was not allowed, so for now, I'm leaving it as two requests.

cdcarter: siiigh thanks for trying

wlonk: Having pared it down to 2 requests, I'm more content to leave it this way than I would have been had it still been the 3 or 4 it was before.

metadeploy/multisalesforce/views.py
```diff
@@ -16,16 +16,49 @@
...
+        # Confirm canModifyAllData:
+        org_info_url = (
+            extra_data["urls"]["rest"] + "connect/organization"
+        ).format(version="44.0")
+        resp = requests.get(org_info_url, headers=headers)
+        resp.raise_for_status()
+        # TODO: Don't use assert outside of tests.
+        assert resp.json()["userSettings"]["canModifyAllData"]
+        # XXX: Also contains resp.json()["name"], but not ["type"], so
+        # insufficient to just call this endpoint.
```
jgerigmeyer: Fix these.

metadeploy/multisalesforce/views.py
```diff
@@ -16,16 +16,49 @@
...
+        headers = {
+            "Authorization": "Bearer {}".format(token),
+        }
+
+        # Confirm canModifyAllData:
+        org_info_url = (
+            extra_data["urls"]["rest"] + "connect/organization"
+        ).format(version="44.0")
+        resp = requests.get(org_info_url, headers=headers)
+        resp.raise_for_status()
```
jgerigmeyer: What will happen here if the user doesn't have permissions? I assume the login will proceed, just without the extra org details (based on the `try...except` below)?

wlonk: Correct. This will raise an `HTTPError` which will get caught below and silently handled. The lack of org data attached to the user is then our indication that they don't have the correct perms.

wlonk:
@jgerigmeyer and @cdcarter, I now feel this is ready to merge, if you can give a review.

PR #104, opened 2018-11-07
[start-install] Start an installation.

jgerigmeyer:
https://trello.com/c/sh6JTbJK

**NOTE:** Running this PR requires first doing the following::

- Add `export DJANGO_HASHID_SALT='something long and random'` to your `$VIRTUAL_ENV/bin/postactivate` script (the value can be any arbitrary string)
- Re-run `workon metadeploy`
- `pip install -Ur requirements/local.txt`
- `./manage.py truncate_data`
- `./manage.py populate_data`
- `./manage.py migrate`
- `bin/unpack-node`
- `yarn`

---


wlonk:
@jgerigmeyer right now, the backend just trusts you to put useful and correct values in the `org_name` and `org_type` fields and doesn't enforce them at all; as they're not used on the backend, and are just for display purposes, do we care about enforcement?

wlonk:
Answer: these don't have to be posted, they are derived from the user at the time of job creation.

jgerigmeyer:
Summarizing access control:

- By default, the job should be private to the creator and is_staff. Anyone else should get a `404` when trying to `GET` the job.
- In the job serializer, only the creator and is_staff should receive `creator.username` and `org_name`. For all others, those should be either `''` or `None`.

metadeploy/api/jobs.py
```diff
@@ -62,6 +62,25 @@ def extract_user_and_repo(gh_url):
     return user, repo
 
 
+@contextlib.contextmanager
+def finalize_result(result):
+    if isinstance(result, PreflightResult):
```
cdcarter: should be able to just use result.Status.failed/result.Status.complete and you'll get the right enum member for the current model

metadeploy/api/jobs.py
```diff
@@ -107,7 +126,7 @@ def zip_file_is_safe(zip_file):
     )
 
 
-def run_flows(user, plan, skip_tasks, flow_class=None, preflight_result=None):
+def run_flows(user, plan, skip_tasks, flow_class=None, result=None):
```
cdcarter: can you docstring or type annotate that result is a models.Model

metadeploy/api/serializers.py
```diff
@@ -110,20 +127,26 @@ class Meta:
 
 
 class JobSerializer(serializers.ModelSerializer):
+    id = serializers.CharField(read_only=True)
     user = serializers.HiddenField(
         default=serializers.CurrentUserDefault(),
     )
+    org_name = serializers.SerializerMethodField(
+        read_only=True,
```
cdcarter: should be readonly without specifying it

metadeploy/api/serializers.py
```diff
@@ -137,14 +160,36 @@ class Meta:
...
         extra_kwargs = {
             'created_at': {'read_only': True},
             'enqueued_at': {'read_only': True},
             'completed_steps': {'read_only': True},
             'job_id': {'read_only': True},
+            'status': {'read_only': True},
+            'org_type': {'read_only': True},
         }
 
+    def requesting_user_has_rights(self):
```
cdcarter: can we dcstring or change this method name to be clear that its rights for some members and not the object itself?

metadeploy/api/serializers.py
```diff
@@ -184,13 +229,16 @@ def validate(self, data):
         )
         if not has_valid_steps:
             raise serializers.ValidationError("Invalid steps for plan.")
+        data["org_name"] = data["user"].org_name
+        data["org_type"] = data["user"].org_type
```
cdcarter: these assignments here are odd to me, but perhaps theres no better place to do it

wlonk: I couldn't find a better place; a `read_only` and `default` on the field seemed most "semantic", but it didn't play well together.

wlonk: And we can be sure that validation will run, so.

metadeploy/api/serializers.py
wlonk: Sure.

metadeploy/api/serializers.py
wlonk: Ah, yes, I think this was an edito, left over from previous expression of this field.

metadeploy/api/jobs.py
wlonk: We aren't using type annotations elsewhere, so I will _at least_ docstring, but a type annotation would be nice too.

metadeploy/api/jobs.py
wlonk: Oh of course! That's a lot nicer; every time I use `isinstance` I feel like a demon gets its wings, y'know?

jgerigmeyer:
@mirisuzanne @stacyk @SondraE In addition to the new job-detail ("run an installation") page -- which is the main addition on this PR -- I wanted to note a few other minor UX updates/additions:

- Job-404 page (if you are logged out):

screenshot 2018-11-19 12 13 59

- Job-404 page (if you're logged in, but don't have permissions to view that past installation):

screenshot 2018-11-19 12 15 19

- Other 404 and error pages now show the same "Desert" illustration (let me know if you prefer a different illustration from https://lightningdesignsystem.com/components/illustration/):

screenshot 2018-11-19 12 16 33
screenshot 2018-11-19 12 16 46
screenshot 2018-11-19 12 16 58
screenshot 2018-11-19 12 17 28
screenshot 2018-11-19 12 17 57
screenshot 2018-11-19 12 18 54

jgerigmeyer:
@mirisuzanne Hm, that's the one example out of all of these that is least likely to occur. It really means "this entire deployment has not been setup properly", in that there are no products in the DB. I'm not sure what action a user could take, other than trying again later or contacting an admin?

src/js/components/404.js
```diff
@@ -2,26 +2,37 @@
...
-          <Link to={routes.home()}>home page</Link>?
-        </p>
-      ) : (
-        children
-      )}
-    </div>
+    <EmptyIllustration
+      message={
+        message === undefined ? (
+          <>
```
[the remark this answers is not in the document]
jgerigmeyer: Ugly? It's what you always wished for in Marionette! :)

In React, a prop cannot be multiple elements, but the `<>...</>` pattern gets around that, basically telling React to treat the children as one element without adding an unnecessary wrapper in the DOM.

https://reactjs.org/docs/fragments.html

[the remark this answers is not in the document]
mirisuzanne: I wished to avoid nesting. I didn't wish for empty tags - but if that's the solution that avoids extra markup, I'll take it. I guess it makes as much code-readability sense as any workaround for that.

src/js/components/authError.js
```diff
@@ -6,18 +6,18 @@ import { Link } from 'react-router-dom';
...
-      <h1 className="slds-text-heading_large">Oh No!</h1>
-      <p>
-        An error occurred while attempting to log in. Try the{' '}
-        <Link to={routes.home()}>home page</Link>?
-      </p>
-    </div>
+    <EmptyIllustration
+      message={
+        <>
+          An error occurred while attempting to log in. Try the{' '}
```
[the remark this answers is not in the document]
jgerigmeyer: Yes and no. JSX (helpfully) ignores linebreak whitespace in text nodes, but when you wrap a sentence you need to tell it to include an extra space. Either:

```jsx
<p>
  This{' '}
  and that.
</p>
```

Or:

```jsx
<p>
  This&nbsp;
  and that.
</p>
```

I used to have the latter, but I think @wlonk thought the former was more legible (and I'm inclined to agree).

mirisuzanne:
> I'm not sure what action a user could take, other than trying again later or contacting an admin?

I think even that is useful information. This isn't your fault, and there's not much you can do. Come back later. If we can alert an admin for you, that might be bonus.

src/js/components/404.js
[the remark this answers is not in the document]
jgerigmeyer: Yeah, the "empty tag" is just an HTML-like shorthand. I don't see much downside.

https://reactjs.org/docs/fragments.html#short-syntax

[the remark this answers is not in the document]
mirisuzanne: ok 👍 approve

src/js/components/authError.js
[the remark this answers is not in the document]
mirisuzanne: Between the two, I agree – but this is exactly the sort of thing I find absurd about JSX – like it was designed without real content in mind. :/

src/js/components/jobs/progressBar.js
```diff
@@ -0,0 +1,54 @@
+// @flow
+
+import * as React from 'react';
+import classNames from 'classnames';
+
+import type { Job as JobType } from 'jobs/reducer';
+
+// Progress Bar has not been implemented yet in design-system-react
+// https://github.com/salesforce/design-system-react/issues/1365
+const ProgressBar = ({ job }: { job: JobType }): React.Node => {
```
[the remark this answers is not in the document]
jgerigmeyer: @mirisuzanne Have you run an installation? That's the main functional change of this PR.

Go to a plan (e.g. http://localhost:8080/products/product-with-useful-data/0.3.1/full-install), log in, run a preflight, then click the "Install" button.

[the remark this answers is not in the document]
mirisuzanne: @jgerigmeyer I hit the "install" button and nothing happened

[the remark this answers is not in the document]
stacyk: Nothing happened for me as well.

src/js/components/jobs/detail.js
```diff
@@ -0,0 +1,234 @@
...
+            </div>
+            <div
+              className="slds-p-around_medium
+                slds-size_1-of-1
+                slds-medium-size_1-of-2"
+            >
+              {(job.creator && job.creator.username) ||
+              job.org_name ||
+              job.org_type ? (
+                <Card
```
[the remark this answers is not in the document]
jgerigmeyer: Sure, will do.

src/js/components/jobs/progressBar.js
```diff
@@ -0,0 +1,54 @@
...
+  const progress = Math.min(
+    Math.round((job.completed_steps.length / job.steps.length) * 100),
+    100,
+  );
+  const id = `${job.id}-progress`;
+  return (
+    <div>
```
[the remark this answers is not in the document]
jgerigmeyer: Hm, I just copied the markup from the "Descriptive Progress Bar" example on https://www.lightningdesignsystem.com/components/progress-bar/

But I don't think the extra `div` is required, so I'll just remove it. 👍

stacyk:
I reviewed this with Mia, so I'll make my review as complete/approved since her questions were documented

jgerigmeyer:
@mirisuzanne @stacyk Okay, I think I addressed all your feedback aside from functional review of the Installation itself.

src/js/components/plans/stepsTable.js
```diff
@@ -246,33 +362,40 @@ const StepsTable = ({
...
   plan: PlanType,
-  preflight: ?PreflightType,
+  preflight?: ?PreflightType,
+  selectedSteps?: SelectedStepsType,
+  job?: JobType,
+  handleStepsChange?: (string, boolean) => void,
 }) => (
-  // DataTable uses step IDs internally to construct unique keys,
-  // and they must be strings (not integers)
   <article className="slds-card slds-scrollable_x">
```
stacyk: This slds-card class on the article is sticking out a pixel above and below the table but removing it removes the left and right border, which is fine with me but maybe not the intended design. Does this double border effect bother anyone? 
slds-card-table-bordered

jgerigmeyer: @stacyk Good catch; I'll fix that.

src/sass/components/_steps.scss
```diff
@@ -2,6 +2,9 @@
 // =====
 
 #plan-steps-table {
+  // Borders are added by parent .slds-card
+  border: none;
```
jgerigmeyer: @stacyk This look okay to you?

stacyk: It works. IF we ever want to get super picky, we can add a `border-top-left-radius: 0.25rem;` to the `.slds-table thead th:first-child` and the same on the right hand side/last-child for small screens to not cut off the corner borders, but that is probably something we won't need.

src/js/components/jobs/detail.js
```diff
@@ -0,0 +1,207 @@
...
+    this.fetchVersionIfMissing();
+    this.fetchJobIfMissing();
+  }
+
+  componentDidUpdate(prevProps) {
+    const { product, version, versionLabel, job, jobId } = this.props;
+    if (
+      product !== prevProps.product ||
+      version !== prevProps.version ||
+      versionLabel !== prevProps.versionLabel
```
wlonk: Wrap this condition in small well-named function. `versionLabelChanged(props, prevProps)` or similar. Ditto below.

src/js/components/jobs/detail.js
```diff
@@ -0,0 +1,207 @@
...
+    const {
+      user,
+      product,
+      version,
+      versionLabel,
+      plan,
+      job,
+      jobId,
+    } = this.props;
+    const blocked = gatekeeper({
```
wlonk: Maybe rename `gatekeeper`, if a better name can be found?

src/js/components/plans/detail.js
```diff
@@ -73,22 +75,96 @@ class PlanDetail extends React.Component<Props> {
...
+      version,
+      versionLabel,
+      user,
+      plan,
+      preflight,
+    } = this.props;
+    if (
+      product !== prevProps.product ||
+      version !== prevProps.version ||
+      versionLabel !== prevProps.versionLabel
```
wlonk: Move this check into a short well-named method, ditto below.

src/js/components/plans/jobResults.js
```diff
@@ -107,43 +106,36 @@ const PreflightResults = ({
...
+    const jobErrors = job.results && job.results.plan;
+    const failed = errorCount > 0 || job.status === CONSTANTS.STATUS.FAILED;
     return (
       <>
         <p className={failed ? 'slds-text-color_error' : ''}>
           {failed ? <ErrorIcon /> : <WarningIcon />}
-          {failed || preflight.is_valid
-            ? `Pre-install validation found ${msg}.`
-            : 'Pre-install validation has expired; please run it again.'}
+          {!failed && job.is_valid === false
```
wlonk: Explanatory comment about the trickiness here plz.

src/js/components/plans/jobResults.js
```diff
@@ -107,43 +106,36 @@ const PreflightResults = ({
...
+        {failed && failMessage ? <p>{failMessage}</p> : null}
+        {jobErrors ? <ErrorsList errorList={jobErrors} /> : null}
       </>
     );
   }
 
-  // Valid preflight without errors/warnings
-  if (preflight.is_valid) {
+  // Invalid job
+  if (job.is_valid === false) {
```
wlonk: Again, tricksiness.

src/js/components/products/detail.js
```diff
@@ -70,6 +69,21 @@ class VersionDetail extends React.Component<VersionDetailProps> {
...
+  componentDidMount() {
+    this.fetchVersionIfMissing();
+  }
+
+  componentDidUpdate(prevProps) {
+    const { product, version, versionLabel } = this.props;
+    if (
+      product !== prevProps.product ||
+      version !== prevProps.version ||
+      versionLabel !== prevProps.versionLabel
```
wlonk: Small well named etc.

src/js/plans/reducer.js
```diff
@@ -60,25 +61,54 @@ const reducer = (
...
+          is_ready: false,
+        },
+      };
+    }
+    case 'PREFLIGHT_COMPLETED': {
+      const preflight = action.payload;
+      const { plan } = preflight;
+      return { ...preflights, [plan]: preflight };
+    }
+    case 'PREFLIGHT_FAILED': {
```
wlonk: C-style switch statement with multiple cases in a row, then the block that applies to all of them.

src/js/utils/api.js
```diff
@@ -46,7 +46,7 @@ const getApiFetch = (onAuthFailure: () => void) => (
           onAuthFailure();
           return getResponse(response);
         }
-        if (response.status === 404) {
+        if (response.status === 403 || response.status === 404) {
```
wlonk: Check where a 403 can come from, adjust backend accordingly.

jgerigmeyer: @wlonk So this `403` comes through when you try to fetch a `job` but don't have permissions (e.g. not logged in). But when you're logged in and try to fetch a `job` that doesn't exist, you get a `404`. Do we want to expose that info publicly (i.e. the difference between "no job" and "there's a job but we can't tell you about it")?

I'm handling them the same on my end, but that's why this change was required.

wlonk: So:

|          | No Job | Job     |
|----------|--------|---------|
| Unauth'd | ?      | 403     |
| Auth'd   | 404    | 404/200 |

Can we answer the question in `?`, too?

I'd like all failures to be 404, not 403, so I'll ensure that.

jgerigmeyer:
@mirisuzanne By "successful install" and "completed installation" do you just mean at least one step has completed? The progress bar does update live, but not until the first step completes successfully...

screenshot 2018-11-19 18 10 33

I'm not sure I understand the problem, but if you or @SondraE or @stacyk has a suggestion for initial "active/loading" styling for the progress bar, I'm fine adding something. I'm just not sure it's needed -- personally my eye is drawn to it as soon as a step completes, which is when I think it's actually helpful information anyway.

src/js/utils/api.js
jgerigmeyer: Right now, unauth'd is returning `403` consistently, and auth'd either `404` or `200`.

|          | No Job | Job     |
|----------|--------|---------|
| Unauth'd | 403      | 403     |
| Auth'd   | 404    | 200 |

mirisuzanne:
@jgerigmeyer yes, I'm talking about completed steps. 

For me, the first step too quite a while, and I wasn't sure if anything was happening at all, or if it had frozen, because there was a bar and no progress. The goal is to give the user a sense that "things are moving forward, just wait". The reason we have loading indicators, progress bars, etc, is because users get antsy and impatient when we aren't sure if there's action – and we *feel* like something moves faster if we see constant progress, even if the final time is exactly the same. So a completed step may be the first actual data-point, but it's not the first time a user needs to hear "hey, we're working, I promise."

src/js/utils/api.js
wlonk: OK. I think I'm maybe OK with that, then; 403 means, in that case, "please send an auth header if you want a chance of seeing this".

metadeploy/api/views.py
```diff
@@ -36,6 +37,13 @@ class JobViewSet(viewsets.ModelViewSet):
         'user',
     )
 
+    def get_queryset(self):
+        if self.request.user.is_staff:
+            return Job.objects.all()
+        return Job.objects.filter(
+            Q(is_public=True) | Q(user=self.request.user),
+        )
+
```
davisagli: Sigh...Django and always having to duplicate access control logic between when you're querying and when you've got an actual object.

src/js/components/authError.js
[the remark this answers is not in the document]
davisagli: Um. You do need this if you want space between elements that are coded on separate lines, but I don't think you need it in the middle of a block of text that spans multiple lines.

src/js/utils/api.js
jgerigmeyer: @wlonk How will that work with `is_public` jobs though?

jgerigmeyer:
@mirisuzanne I get that. I kinda thought the two spinners on the page were enough to show "something is happening... please wait." But I'm open to changing the progress bar too -- do you have a suggestion for what would be better?

src/js/components/authError.js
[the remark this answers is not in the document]
jgerigmeyer: @davisagli Yes, thanks for correcting my inaccurate contrived example.

This works:

```jsx
<p>
  This
  and that.
</p>
```

But this is where it's needed:

```jsx
<p>
  This{' '}
  <a>and that</a>.
</p>
```

src/js/utils/api.js
wlonk: It won't, fixing.

wlonk: Fixed with test coverage for the question.

jgerigmeyer:
Ok... I think this is ready for merge, pending possibly a suggestion for progress-bar from @mirisuzanne? (Though that could happen in a separate branch anyway.)

mirisuzanne:
@jgerigmeyer let's not hold up merge – I think it will require conversations with @SondraE about UX solutions, and with you about what's possible.

metadeploy/api/views.py
wlonk: No great answer to it, though https://schinckel.net/2018/11/12/django-computedfield%28%29/ is promising.

wlonk:
@jgerigmeyer 🚀

PR #105, opened 2018-11-07
📦 🛢 [hashids] Use Hash IDs with every public-facing ID

wlonk:
baby gecko


metadeploy/api/models.py
```diff
@@ -16,6 +16,7 @@
 from asgiref.sync import async_to_sync
 from colorfield.fields import ColorField
 from model_utils import Choices, FieldTracker
+from hashid_field import HashidAutoField
```
cdcarter: oh hey I didn't know about this django implementation, awesome!

wlonk: It's pretty good; there's a little mishegas to get it to play well with DRF, but not too bad.

env.example
```diff
@@ -11,6 +11,7 @@ export POSTGRES_DB=metadeploy
 # General settings
 export DJANGO_SETTINGS_MODULE=config.settings.local
 export DJANGO_SECRET_KEY='CHANGEME!!!'
+export DJANGO_HASHID_SALT='something long and random'
```
jgerigmeyer: If/when this gets backported to project-template, we'll have to update the post-hook to replace this string.

wlonk: I already changed the `app.json`, but didn't think about the cookiecutter post-hook part.

jgerigmeyer: I went ahead and made the project-template changes locally; I'll push when we merge this branch and I backport it.

PR #107, opened 2018-11-08
[change-channel-layers] Send messages to groups named for models, not users

wlonk:
red panda


wlonk:
OK, @jgerigmeyer, we're gonna do what you suggested: when you connect to the websockets at `ws/notifications/`, send a follow-up bit of JSON to subscribe to specific model updates:

```
{
    "model": // the model name, all lower case. "user", "preflightresult", "job", etc.
    "id": // model ID.
}
```

wlonk:
It occurs to me you don't have user IDs, so I'll add that in here.

wlonk:
… that's all in `start-install`, isn't it.

metadeploy/api/push.py
```diff
@@ -1,10 +1,27 @@
...
+"""
+
+
 from channels.layers import get_channel_layer
 
 
-async def push_message_to_user(user, json_message):
-    user_id = user.id
+async def push_message_about_instance(instance, json_message):
+    model_name = instance.__class__.__name__.lower()
```
cdcarter: `instance._meta.model_name` (assuming theres no reason to use the app name too/ctype label)

wlonk: Mm, that's better, yeah.

metadeploy/consumers.py
```diff
@@ -7,21 +7,28 @@ class PushNotificationConsumer(AsyncJsonWebsocketConsumer):
...
+                group_name,
+                self.channel_name,
+            )
+        self.groups = set()
 
     async def notify(self, event):
         await self.send_json(event['content'])
+
+    async def receive_json(self, content, **kwargs):
+        # Just used to subscribe to notification channels.
```
cdcarter: is there reasonable benefit to making the payload something like
```
{'action':'subscribe', 'model': 'plan', 'id': 'abc123'}
```
in the event we need to support more two way communications? not worth it at this time is a very valid response.

also, yes, definitely confirm access rights here. I'd love to see it done in a semi-idealistic-way with django's object level perms, even if we don't go full on django-guardian. but that's also very flexible if its too much work.

wlonk: Here's my thought process there: if I could see the scenario where we'd push from the client to the websocket, I'd add that for sure—it's more explicit. But as I can't see that scenario on any horizon yet, I won't, because I can't reasonably determine what the needs and constraints might be.

Yeah, I think that Django's object level perms is probably a good route here. I'm unhappy that this might touch the DB, but I think it's necessary here.

wlonk: I'm afraid that Django still seems to be all about the table-level perms; I'm gonna look at `django-guardian` though, as it seems the best solution.

metadeploy/consumers.py
```diff
@@ -25,10 +27,28 @@ class PushNotificationConsumer(AsyncJsonWebsocketConsumer):
...
+        if not self.has_good_permissions(content):
+            return
+        group_name = f"{content['model']}-{content['id']}"
+        self.groups.add(group_name)
+        await self.channel_layer.group_add(
+            group_name,
+            self.channel_name,
+        )
+
+    def has_good_permissions(self, content):
```
cdcarter: this method is doing two things, is the model subscribeable, and does the user have the permissions to subscribe to it.

jgerigmeyer:
@wlonk Reading through the code, it looks like there's still more work to be done on your end here?

wlonk:
@jgerigmeyer yep.

wlonk:
@jgerigmeyer I think what remains on this is frontend work. I'd love to get this merged early this week if possible?

metadeploy/consumers.py
```diff
@@ -1,21 +1,76 @@
...
+        if content["model"] == "job":
+            # If we use this version, we must make this an async
+            # function.
+            # from channels.db import database_sync_to_async
+            # job = await database_sync_to_async(self.get_job)(content["id"])
+            # Why does database_sync_to_async not work in the tests?
+            job = self.get_job(content["id"])
+            return job and job.visible_to(self.scope["user"])
+        return True
```
jgerigmeyer: @wlonk Hm... shouldn't we be checking permissions for `user` and `preflightresult` objects too? It seems odd to me that anyone could subscribe to events on any user or for any preflight.

metadeploy/consumers.py
```diff
@@ -1,21 +1,87 @@
...
+
+
+def user_context(user):
+    return {"request": Request(user)}
 
 
 class PushNotificationConsumer(AsyncJsonWebsocketConsumer):
     async def connect(self):
         if self.scope["user"].is_anonymous:
             await self.close()
```
jgerigmeyer: We now need to accept anonymous connections.

wlonk:
@jgerigmeyer this should address all the issues in the ticket:

 - Creating a preflight returns the preflight.
 - All data can be passed across the redis layer correctly now (though it would be great to get robust patterns for this going forward—a recurring issue is that the FieldTracker breaks transparent model serialization, and that Serializer subclasses [not instances, but classes] can't be serialized across the layer either).
 - The `BACKEND_ERROR` without `JOB_FAILED` was because the error was in sending the Job data across the redis layer to the consumer in the first place 😖 
 - Anonymous connections are possible now, which is OK because we are more rigorous about checking whether you're allowed to subscribe to any object you subscribe to.
 - Also, bonus round: sending the subscribe packet gets a response with an `"ok"` or `"error"` key as appropriate.

metadeploy/api/models.py
```diff
@@ -569,8 +588,11 @@ def push_if_invalidated(self):
...
         ret = super().save(*args, **kwargs)
 
-        self.push_if_completed()
-        self.push_if_failed()
-        self.push_if_invalidated()
+        try:
+            self.push_if_completed()
+            self.push_if_failed()
+            self.push_if_invalidated()
+        except RuntimeError as error:
```
cdcarter: whats the error happening here? just websocket push failures?

wlonk: It's actually a transient test error 😢

src/js/utils/websockets.js
```diff
@@ -90,12 +104,21 @@ export const createSocket = ({
...
+  const socket = new Sockette(url, {
     protocols: opts.protocols,
     timeout: opts.timeout,
     maxAttempts: opts.maxAttempts,
     onopen: e => {
       log('[WebSocket] connected');
+      open = true;
+      for (const obj of pending) {
+        log('[WebSocket] subscribing to:', obj);
+        socket.json(obj);
```
wlonk: Rename `obj` to `payload` for consistency with below?

metadeploy/api/serializers.py
```diff
@@ -272,7 +275,9 @@ def validate(self, data):
 
 
 class PreflightResultSerializer(ErrorWarningCountMixin, serializers.ModelSerializer):
+    id = serializers.CharField(read_only=True)
     plan = IdOnlyField(read_only=True)
+    user = IdOnlyField(read_only=True)
     is_ready = serializers.SerializerMethodField()
     error_count = serializers.SerializerMethodField()
     warning_count = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
wlonk: Historically, I've had trouble declaring serializer fields on a mixin due to metaclass shenanigans. I can try again and see if that's an old outdated habit though?

metadeploy/api/views.py
```diff
@@ -58,8 +58,12 @@ def preflight_get(self, request):
 
     def preflight_post(self, request):
         plan = self.get_object()
-        preflight_job.delay(request.user, plan)
-        return Response("", status=status.HTTP_202_ACCEPTED)
+        preflight_result = PreflightResult.objects.create(
+            user=request.user, plan=plan, organization_url=request.user.instance_url
+        )
+        preflight_job.delay(preflight_result.pk)
```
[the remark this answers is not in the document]
wlonk: Sub.

metadeploy/api/serializers.py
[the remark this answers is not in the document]
cdcarter: I think you may need to declare Serializer as a base on the mixin? But it's really nbd, feel free to leave it.

metadeploy/api/views.py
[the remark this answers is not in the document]
jgerigmeyer: @cdcarter Is this blocking merge? Or just something for further conversation?

PR #115, opened 2018-11-20
[prominent-productname] Emphasize Product Name

stacyk:
## Trello Card
- https://trello.com/c/WVUeE6qS/73-as-an-admin-i-want-the-product-version-to-be-more-prominent-on-plan-detail-and-job-detail-pages

## What to look for
On the [Plan Details](http://localhost:8080/products/product-with-useful-data/0.3.1/full-install) and Job Details (after install) pages, we are emphasizing the Product Name and version of the current plan.


jgerigmeyer:
@mirisuzanne @SondraE Here are screenshots showing what the new section looks like on both plan-detail and job-detail:

screenshot 2018-11-21 14 46 29

---

screenshot 2018-11-21 14 47 42

SondraE:
@jgerigmeyer 
Am I understanding correctly that "view available plans" link and "Product With Useful..." link in the header both go to the product detail page which has the list of plans? The concern I heard was just that the name of the product was getting lost. Is there an issue with the link too? If we do want that link twice, is there a problem with making the name of the product the link? Making it a blue link would perhaps make it stand out even more. 

Could we put "This plan is part of..." right under "Full Install" as a subtitle? My concern is that the generic text could be quite long and push the product name way down the screen. In my mind, it makes sense to group that info with the title.

Hmm... As I think about this, I'm just not sure adding the product title and link to plans to the instruction text area makes it more accessible. There are times when there can be several paragraphs of text in that area as well as warnings and errors. Seems like it could get cluttered fast. Maybe it would be worthwhile to rethink the page header to make the link back to plans as well as the product title more obvious.

jgerigmeyer:
@SondraE Okay... we can discuss further with @mirisuzanne and @stacyk on Monday, in that case. 👍

PR #116, opened 2018-11-21
🐈 [preflight-warnings] Require confirm before install with warnings.

jgerigmeyer:
https://trello.com/c/dzQ55CDO


src/js/components/plans/ctaButton.js
```diff
@@ -63,17 +76,39 @@ export const ActionBtn = ({
...
+    doStartJob({ plan: plan.id, steps: [...selectedSteps] }).then(action => {
+      const { type, payload } = action;
+      if (type === 'JOB_STARTED' && payload && payload.id) {
+        const url = routes.job_detail(
+          productSlug,
+          versionLabel,
+          plan.slug,
+          payload.id,
+        );
+        history.push(url);
```
davisagli: Is there anything preventing the user from navigating back and starting the job again?

jgerigmeyer: No, but I think that's by design? Or at least I haven't heard of any restrictions desired on that as of yet.

davisagli: Well, we definitely don't want to let users run more than one concurrent flow on the same org. We should add a story and discuss it if that's new.

jgerigmeyer:
@mirisuzanne @SondraE I'd appreciate functional/style review on this. Here's what it looks like with one warning:

screenshot 2018-11-21 12 11 36

And here's with multiple warnings:

screenshot 2018-11-21 12 14 55

SondraE:
@jgerigmeyer It's a little confusing with the meta text examples, but I'm not seeing anything obvious that I would change about the UX.

mirisuzanne:
@jgerigmeyer my thoughts:

- header could be more clear, something like: "Potential Issues (confirm to continue)"
- we could group warnings/confirmations by step, rather than repeating the step name
- we could also use the warnings as checkbox-labels, rather than repeating "I understand"

```
# Potential Issues (confirm to continue)
[ ] This is a global warning
[ ] This is another global warning

## Step1 Name
[ ] This is a step1 warning
[ ] This is another step1 warning

## Step2 Name
[ ] This is a step2 warning
[ ] This is another step2 warning
```

jgerigmeyer:
@mirisuzanne I like that direction. One concern: warning messages can contain arbitrary HTML (e.g. links)... I'm not sure how that would work within a checkbox label?

jgerigmeyer:
@mirisuzanne Revised:

screenshot 2018-11-26 14 42 13

PR #117, opened 2018-11-26
🛢 [post-install] Add post-install message to JOB_COMPLETED WS event

wlonk:
painted puppies


cdcarter:
Two thoughts, intertwined:
- preflight_message should also be bleached markdown
- post_install_message should also be a TextField with no arbitrary max_length

wlonk:
@cdcarter Totally agree. The gist you link (I've finally gotten to look over it) looks pretty good; I might want to make sure it JSON serializes as-expected, though, for REST API purposes. Do we want to put that in the template or just a separate package that we'll depend on? (I lean towards the latter, but it's putting a maintenance burden on you in a way, so I want to check.)

cdcarter:
I also lean toward the latter. The exact strategy (include the logfmt utils? rq utils? all of multisalesforce? or make many packages ((upfront yuck))) I think we can defer a few days.

metadeploy/api/push.py
```diff
@@ -92,7 +92,10 @@ def user_context(user):
     from .serializers import JobSerializer
 
     user = job.user
-    payload = JobSerializer(instance=job, context=user_context(user)).data
+    payload = {
+        'job': JobSerializer(instance=job, context=user_context(user)).data,
+        'message': job.plan.post_install_message_markdown,
+    }
```
jgerigmeyer: @wlonk I would expect this to be handled the same as `preflight_message` -- that is, included in the `plan` serializer instead of returned separately in the WS message. Otherwise it's not accessible on page-reload with a simple `GET`, no?

wlonk: @jgerigmeyer done. It's included in the job serializer regardless, but you can/should show it based on the job's status, right?

jgerigmeyer: Yes, that's what I was thinking.

metadeploy/api/models.py
```diff
@@ -340,9 +343,37 @@ class Plan(HashIdMixin, SlugMixin, models.Model):
...
+            attributes=self.markdown_attrs,
+        )
+
+    @property
+    def post_install_message_markdown(self):
+        return bleach.clean(
+            markdown(self.post_install_message),
+            tags=self.markdown_tags,
+            attributes=self.markdown_attrs,
+        )
```
jgerigmeyer: This reminds me... I see that `skip` and `optional` messages are also getting cleaned with bleach, but on the front-end we're not inserting those as unescaped HTML. @cdcarter Do we anticipate markdown or links or markup of any kind in `skip` and `optional` preflight result messages?

PR #120, opened 2018-11-27
📦 Use VCR for better test of happy path on jobs

wlonk:
red panda


wlonk:
Note the `REDACTED` value for the GitHub tokens. Be careful when adding new VCR tests not to leak tokens.

PR #127, opened 2018-12-04
[order-categories] 🛢 Order products and categories

wlonk:
Simple manual ordering on the admin, setting the ordering Meta value should keep them ordered that way when the client requests them.

See https://trello.com/c/sbfB4jl2/51-as-salesforce-staff-i-want-to-control-the-displayed-ordering-of-categories-and-products-within-a-category

snow pup pony


wlonk:
@jgerigmeyer That change sounds good and coherent, I will do it.

wlonk:
@jgerigmeyer If you can re-review this when you get a chance, I think it's working now.

requirements/base.txt
```diff
@@ -22,7 +22,7 @@ Pillow==5.3.0
 django-model-utils==3.1.2
 django-storages==1.7.1
 boto3==1.9.62
-github3.py==0.9.6  # pyup: ignore
+github3.py==1.2.0
```
wlonk: Why is this safe now?

jgerigmeyer: As of cumulusci==2.2.0: https://github.com/SFDO-Tooling/CumulusCI/blob/master/requirements.txt#L10

wlonk: Nice!

davisagli: We updated cumulusci to require github3.py 1.2.0. If metadeploy is using github3.py directly you might need to update things. If you can point me to where it's used I can look to see if there are obvious changes needed.

jgerigmeyer: @davisagli It's only used one place (https://github.com/SFDO-Tooling/metadeploy/blob/master/metadeploy/api/jobs.py#L161), and we were actually using v2.2.0 before pinning it back because cumulusci was complaining (https://github.com/SFDO-Tooling/metadeploy/commit/1a8162791e2a9aa34780475d16bd1ee623115285). So I think we're good to go?

PR #130, opened 2018-12-05
[handle-dyno-death] 🛢 Handle SIGTERM by marking a job as canceled

wlonk:
This relies on the exception propagated on SIGTERM from RQ. If a job manages to finish in the 30s before Heroku sends the followup SIGKILL, it should get a completed status, overriding the canceled status.

See https://trello.com/c/PJcbJsi7/35-as-an-admin-i-want-to-not-be-left-hanging-if-heroku-me-screws-up

ozzie


metadeploy/api/jobs.py
```diff
@@ -132,6 +143,7 @@ def run_flows(
 
     with contextlib.ExitStack() as stack:
         stack.enter_context(finalize_result(result))
+        stack.enter_context(mark_canceled(result))
```
cdcarter: this makes me happy!

metadeploy/api/jobs.py
```diff
@@ -83,6 +84,16 @@ def prepend_python_path(path):
         sys.path = prev_path
 
 
+@contextlib.contextmanager
+def mark_canceled(result):
```
cdcarter: please add a docstring explaining the logic behind this, i.e. sigterm will cause the rq worker to raise that StopRequested?

if you're not paying attention it looks a LOT like you're handling StopIteration on a generator

PR #133, opened 2018-12-05
[flush-credentials] Exclude users with uninfished jobs from token expiry

wlonk:
See https://trello.com/c/5esBWARK/49-as-an-admin-i-dont-want-credentials-for-my-org-to-persist-longer-than-they-have-to

pupper


wlonk:
@cdcarter We're already scanning for old tokens and removing them, but this should make sure we don't delete ones where there are pending jobs.

PR #134, opened 2018-12-06
🐈 Upgrade deps and hopefully fix CI.

jgerigmeyer:
So my working theory is that the build was failing because the various linting tasks were running in parallel and taking too long. This changes to run them serially.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -22,8 +22,8 @@ settings:
             - static
             - node_modules
   react:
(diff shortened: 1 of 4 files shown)


wlonk:
@jgerigmeyer I looked and looked and finally saw that down in `package.json` one of the lines wasn't a dep upgrade but a change to a script. Not my favorite pattern! But ultimately legible.

PR #138, opened 2018-12-06
Skip running flow on CI for now

jgerigmeyer:
https://github.com/flowtype/flow-bin/issues/138


.circleci/config.yml
```diff
@@ -83,6 +83,11 @@ jobs:
           command: |
             . venv/bin/activate
             yarn lint
+            # @@@ We'd like to run `flow` here, but:
+            # https://github.com/flowtype/flow-bin/issues/138
+    #   - store_artifacts:
+    #       path: /tmp/flow
+    #       destination: flow-logs
```
wlonk: Can we remove instead of comment out this block?

jgerigmeyer: We certainly could... I was leaving it there to remind me that I'd like to reactivate it once we're using Flow again on CI, at least for a while to be sure it's fixed.

PR #139, opened 2018-12-06
[adminapi] Admin API

cdcarter:
Adds an internal API under the admin url namespace (and environment variable) so that CRUD against things like `http://testserver/admin/rest/plans/{plan_id}` works roughly how you'd expect. There's a clickable(ish) in browser API available to superusers who have an active session, and authtokens can be generated in the admin or via manage.py for the upcoming CCI task client.

Nice to have that didn't happen here:
- generated form work nice for ArrayField as well as for Steps
- refactor URL reversing
- swagger/options response/general metadata?
- filtering


metadeploy/adminapi/urls.py
```diff
@@ -0,0 +1,28 @@
...
+    PlanViewSet,
+    ProductCategoryViewSet,
+    ProductSlugViewSet,
+    ProductViewSet,
+    VersionViewSet,
+)
+
+
+def _get_api_basename(viewset):
+    model_name = getattr(viewset, "model_name", None)
```
wlonk: You rely on this being a string, so don't add a `None` fallback here.

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def get_serializer_class(self):
+        self.serlializer_base.Meta.model = self.model
+        return self.serlializer_base
+
+    def get_serializer_context(self,):
+        ctx = super().get_serializer_context()
+        ctx[
+            "route_ns"
+        ] = self.route_ns  # add the route namespace to the serializer context
```
wlonk: Move the comment to the line above, to avoid Black doing this weirdness.

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,83 @@
...
+        return apps.get_model(
+            app_label=str(self.model_app_label), model_name=str(self.model_name)
+        )
+
+    def get_queryset(self):
+        model = self.model
+        return model.objects.all()
+
+    def get_serializer_class(self):
+        self.serlializer_base.Meta.model = self.model
```
wlonk: I'd kinda rather something that's like… I dunno… `self._make_serializer_for(model)`, which will copy the class and modify it, and gets called in the view setup. I don't like that this mutates the `AdminAPISerializer.Meta` class; I suspect that will lead to bad surprises down the road.

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,83 @@
...
+        if "view_name" in field_kwargs:
+            # we're in a hyperlinkedrelationshipfield, need to fix the view ref...
+            field_kwargs["view_name"] = view
+        return field_class, field_kwargs
+
+
+class AdminAPIViewSet(viewsets.ModelViewSet):
+    model_app_label = "api"
+    model_name = None
+    permission_classes = [permissions.IsAdminUser]
```
wlonk: This line is _vital_, and might be worth emphasizing with a comment.

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,83 @@
...
+    serlializer_base = AdminAPISerializer
+    route_ns = "admin_rest"
```
wlonk: This feels a bit un-DRY. Can we derive this value here, or in the URLs?

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,83 @@
...
+
+class AdminAPISerializer(serializers.HyperlinkedModelSerializer):
+    id = serializers.CharField(read_only=True)
+
+    class Meta:
+        model = None
+        fields = "__all__"
+
+    def build_url_field(self, field_name, model_class):
+        view = f"{self.context['route_ns']}:{self.Meta.model._meta.object_name.lower()}-detail"  # noqa
```
wlonk: Rather than `noqa`-ing a long string like this, I do like to break it out, unless Black is undoing that.

```
view = (
    f"{self.context['route_ns']}:"
    f"{self.Meta.model._meta.object_name.lower()}-"
    f"detail"
)
```

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,83 @@
...
+        field_kwargs = {"view_name": view}  # override default view_name
+
+        return self.serializer_url_field, field_kwargs
+
+    def build_relational_field(self, field_name, relation_info):
+        field_class, field_kwargs = super().build_relational_field(
+            field_name, relation_info
+        )
+        model_field, related_model, to_many, to_field, has_through_model, reverse = (
```
wlonk: You can probably replace some of these with `_` for the linter's sake?

metadeploy/adminapi/api.py
cdcarter: yea, I need to untangle router/viewset/hyperlinkedrelationfield a little bit, and ideally rely on the app_name in urls.py?

cdcarter: I'm not sure the right place for this value to end up, i guess. but I agree it needs to be better than it is now.

metadeploy/adminapi/api.py
cdcarter: agreed, and I think we'll also want to IP restrict these routes

metadeploy/adminapi/api.py
cdcarter: I think it's actually a namedtuple, I should be able to just get related_model.

metadeploy/adminapi/api.py
wlonk: Do you propose doing that in the app logic, or can we do it at the infrastructure level?

metadeploy/adminapi/api.py
cdcarter: I pushed a slightly different approach, still staying within the normal call chain for GenericAPIView.

metadeploy/adminapi/api.py
cdcarter: on the common runtime it has to be in app logic, I can see if private spaces let us do it in infra...

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,91 @@
...
+    def get_queryset(self):
+        model = self.model
+        return model.objects.all()
+
+    def get_serializer_class(self):
+        class AdminSerializer(self.serlializer_base):
+            class Meta(self.serlializer_base.Meta):
+                model = self.model
+
+        return AdminSerializer
```
wlonk: Well, this has the tradeoff of class creation at each call; I don't know that that'll be bad, but it might be worth testing if that happens and causes any appreciable problem?

metadeploy/adminapi/tests.py
```diff
@@ -0,0 +1 @@
+pass
```
wlonk: This could just be empty for now?

metadeploy/adminapi/api.py
cdcarter: Based on my limited test, it seemed nontrivial to deepcopy the Serializer definition itself, but I can dig in a little further. Moving when its instantiated isn't too difficult, I suspect if I try to make Swagger describe work, it'll be useful to do it at app startup instead of request startup anyway...

cdcarter: Oh I did actually get this pattern from DRF, not that that makes it necessarily a good idea... https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L1606

wlonk: I realize I complained when you did it one way, then complained when you did the the other. I really am OK with this approach, I just want to make sure it's not got hidden costs. I expect it won't!

wlonk:
@cdcarter I like the direction this is headed, I'll review and probably 👍 when you say it's ready for that.

metadeploy/adminapi/api.py
davisagli: I adjusted this to memoize as self.serializer_class so it's only constructed once.

metadeploy/adminapi/tests/api.py
```diff
@@ -0,0 +1,52 @@
...
+class TestPlanViewSet:
+    def test_list(self, staff_client, plan_factory):
+        plan = plan_factory()
+
+        url = "http://testserver/admin/rest/plans"
+        response = staff_client.get(url)
+
+        assert response.status_code == 200
+        version_url = f"http://testserver/admin/rest/versions/{plan.version.id}"
+        assert response.json() == {
```
cdcarter: philosophically, I go back and forth on this style of asserting the response...

this test provides an interesting check, if you add a new column on the model, the test will fail if you don't add a new member property here. this is unintuitive, because this is a test of the API layer, not the model, and as implemented, the API layer isn't affecteed by model changes.

There's an argument to be made about breaking public API contracts by...adding a new field? I don't fully buy it but I know it exists in codegen situations and stuff. But in any case, I think that that sort of "api contract" or (as core calls them) "goldfile" testing is better done by diffing swagger....

instead, here, i'd love to see an assert that the response has data, links, and meta members, that the data member is an array of objects, and that object has an "id" key, and a "url" key that matches the ID key...

but, having tests is better than not having tests!

metadeploy/adminapi/api.py
```diff
@@ -1,10 +1,22 @@
...
 from rest_framework.response import Response
 
 
+class IsAllowedIPAddress(permissions.BasePermission):
+    """Permission check for allowed IP networks.
+    """
+
+    def has_permission(self, request, view):
+        ip_addr = IPv4Address(request.META["REMOTE_ADDR"])
+        return any(ip_addr in subnet for subnet in settings.ADMIN_API_ALLOWED_SUBNETS)
```
cdcarter: instead of just returning false which will 403 the view -- actually raise django.core.exceptions.SuspiciousSession, which will 403 the view but also provide a specifically auditable log line

config/settings/base.py
```diff
@@ -180,6 +181,12 @@ def env(name, default=NoDefaultValue, type_=str):
 # Must end in a /, or you will experience surprises:
 ADMIN_AREA_PREFIX = "admin/"
 
+ADMIN_API_ALLOWED_SUBNETS = env(
+    "ADMIN_API_ALLOWED_SUBNETS",
+    default="127.0.0.1/32",
+    type_=lambda x: [IPv4Network(s) for s in x.split(",")],
```
cdcarter: i'd love to get a test in place where this is multiple CIDR blocks to confirm there doesn't end up being any issues with the dots and slashes and strings

metadeploy/adminapi/api.py
davisagli: We talked about this and are going to leave it as is for now. The ways of avoiding the duplication feel like more complexity than is needed until this becomes a generic solution.

davisagli:
@cdcarter @wlonk ready for final review I think

cdcarter:
I just made a commit in the browser, because I just needed to express the ideas ... I can clean it up if y'all approve or if David wants to take a crack... The TODOs are because I could only edit the one file at a time.

I removed some addressed TODOs and some aspirational ones.

I changed the way we're constructing the IP Addr permissions, so that we could limit different viewsets to different ranges. I think this is a good thing to put in template helpers. I think most of the admin api is good to put in template helpers. This is the part I expect to get push back on, because it's certainly a little bit "wha" when yo first look at it, but the type comments help (in my mind). 

Finally, I added TokenAuthentication to the AdminAPI (and a todo to add the installed_app), so that we can mint auth tokens for the cci client. 

Along those lines -- I also added a TODO I'd like @wlonk's feedback on, which is setting (TokenAuthentication,SessionAuthentication) as the default auth settings (like seen in https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme). It's my understanding that rest_framework wants a default or per view authentication class in addition to the authentication backend set for django as a whole. I believe that SessionAuthentication does additional (or any??) CSRF checking... I need to get my local build up again so I can drop a breakpoint and see what's going on in the authentication callstack...

config/settings/base.py
```diff
@@ -211,6 +222,13 @@ def env(name, default=NoDefaultValue, type_=str):
 )
 SECURE_HSTS_PRELOAD = env("SECURE_HSTS_PRELOAD", default=False, type_=boolish)
 
+REST_FRAMEWORK = {
+    "DEFAULT_AUTHENTICATION_CLASSES": (
+        "rest_framework.authentication.TokenAuthentication",
+        "rest_framework.authentication.SessionAuthentication",
+    )
+}
```
davisagli: @cdcarter @wlonk will this interfere at all with the non-admin api?

metadeploy/adminapi/api.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def AllowedIPRange(
+    ip_ranges: List[IPv4Network], cls_prefix: str = ""
+) -> Type[permissions.BasePermission]:
+    """Factory that returns an IsAllowedIpAddress permission for a list of ip_ranges.
+    """
+    return type(
+        f"{cls_prefix}AllowedIPRange", (IsAllowedIPAddress,), {"ip_ranges": ip_ranges}
+    )
```
davisagli: I would have preferred to wait and do this generalization once we needed it in more than one place, particularly since I try to avoid class generation unless there's a _compelling_ reason since it's hard to read and hard to debug. But since it's here I guess it can stay.

config/settings/base.py
cdcarter: the rest api defaults we inherited before this specified SessionAuthentication and BasicAuthentication. We do not want BasicAuthentication. being able to do tokens against the API seems handy for testing, knowing that we control token issue. So, no, IMO this only increases our security posture

metadeploy/adminapi/api.py
cdcarter: avoiding class generation is sadly at odds with using rest_framework's quasi-declarative infrastructure. I prefer this to stacks on stacks of decorators. but this truly is de gustibus. I like this because it makes the declaration in the viewset have (IMO) clearer intent. 

I give you permission to back it out in a month ;)

config/settings/base.py
davisagli: Probably not for this PR, but it looks like we need to add permission checks to the viewsets in the non-admin API?

cdcarter: that's what not merging master into the latest branch will get ya! especially when theres in-practice conflicts that don't appear problematic to a diff!

davisagli: Oh cool, there it is. Yeah, I was looking at this branch locally thinking it was master.

config/settings/base.py
```diff
@@ -115,6 +120,7 @@ def env(name, default=NoDefaultValue, type_=str):
     "allauth.socialaccount",
     "colorfield",
     "rest_framework",
+    "rest_framework.authtoken",
```
wlonk: Adding this app will require running migrations; can you add 🛢 to the PR title to communicate that?

metadeploy/adminapi/api.py
wlonk: I'm pretty sympathetic to @davisagli's point here, but I share the "since it's here…" sentiment.

metadeploy/adminapi/urls.py
```diff
@@ -9,16 +10,15 @@
...
-# TODO: Schema, create a schema endpoint
-
 
 def _get_api_basename(viewset):
     return viewset.model_name.lower()
 
 
 app_name = "admin_api"
-router = routers.DefaultRouter()
+router = routers.DefaultRouter(trailing_slash=False)
```
wlonk: Why no trailing slash? Purely an aesthetic consideration?

config/settings/base.py
cdcarter: oops, I thought the robot did that for me. yes.

metadeploy/adminapi/urls.py
cdcarter: yea, we default to following the heroku api styleguide (and no trailing slashes was a railism that they stuck with)

cdcarter:
I'm going to merge this and backlog splitting up the module, with the intent of doing it in a way thats meaningful for the template...

config/settings/base.py
jgerigmeyer: The robot adds it to the commit message, not to the PR title.

jgerigmeyer:
@cdcarter @wlonk Should this get back-ported to the sfdo-template?

PR #142, opened 2018-12-07
[delist] 🛢 Add is_listed boolean to Product/Plan/Version

wlonk:
See https://trello.com/c/fqPURZJ1/81-1-delist-a-product-version-and-or-plan

This merely marks the models to be omitted in the frontend; they are still present in the listing so that they can easily be reached by URL directly.

skwerl


metadeploy/api/models.py
```diff
@@ -256,6 +257,7 @@ class Version(HashIdMixin, models.Model):
         default="master",
         help_text="This is usually a tag, sometimes a branch.",
     )
+    is_listed = models.BooleanField(default=True)
```
jgerigmeyer: @wlonk I think `is_listed=False` should prevent a version from being the `most_recent_version`.

jgerigmeyer: That also raises the question of what to do with a product that has no versions `is_listed`. I suppose we treat that as `is_listed=False`?

PR #143, opened 2018-12-07
[concurrent-jobs] Disallow concurrent jobs

wlonk:
See https://trello.com/c/xqvUhcAk/78-3-as-an-admin-i-want-to-be-prevented-from-running-concurrent-jobs-on-the-same-org

~~We'll eventually need to add support for listing pending jobs and cancelling them, but for now, this just blocks creation of a new job while one is pending.~~

~~This also uses a user as a proxy for an org, which may not always be correct.~~

puppers


metadeploy/api/flows.py
```diff
@@ -25,6 +30,16 @@ def _get_step_id(self, task_name):
...
 
+    def _pre_task(self, task):
+        """
+        Before each task, we should check if we've been told to abandon this job.
+        """
+        if self._flow_canceled():
+            raise StopFlowException("Job canceled.")
+
+    def _flow_canceled(self):
+        return bool(cache.delete(f"CANCEL-{self.result.id}"))
```
cdcarter: hoist the prefix into a constant, somewhere. probably also make it `metadeploy:cancel:{jobid}`-ish, to look like rq keys...

and i think we should DEFINITELY gert a log line somewhere in here

cdcarter: I also might consider just checking that the cancel key exists, and not clearing it in this otherwise non-state-modifying situation. cache.get() lets you set a timeout...ithink...?

wlonk: Yeah, I think it does.

metadeploy/api/views.py
```diff
@@ -29,6 +30,9 @@ def get_queryset(self):
             filters = Q(is_public=True) | Q(user=self.request.user)
         return Job.objects.filter(filters)
 
+    def perform_destroy(self, instance):
```
cdcarter: wondering if we should stamp a "cancel_request" date/who onto the record too...

wlonk: I think that the timestamp is well worth it. As to the "who", perhaps we should just limit it to the owner for now?

wlonk:
Per Trello, this just requires wiring up the frontend now.

jgerigmeyer:
@wlonk Can you merge `master` into this PR when you get a chance?

metadeploy/api/serializers.py
```diff
@@ -329,6 +329,11 @@ def _get_from_data_or_instance(self, data, name, default=None):
             return value.all()
         return value
 
+    def _pending_job_exists(self, *, plan, user):
+        return Job.objects.filter(
+            status=Job.Status.started, plan=plan, user=user
```
jgerigmeyer: Shouldn't this be filtering by `org`, not `user`?

Also, I don't think `plan` is relevant here?

wlonk:
@jgerigmeyer I believe only frontend work remains on this.

metadeploy/api/push.py
```diff
@@ -102,3 +106,32 @@
...
         type_ = "JOB_CANCELED"
     await push_serializable(job, JobSerializer, type_)
+
+
+async def _job_or_preflight_changed(org_url, type_):
+    from .serializers import OrgSerializer
+    from .models import Job, PreflightResult
+
+    current_job = Job.objects.filter(
+        organization_url=org_url, status=Job.Status.started
```
cdcarter: sorry for just catching this now...

Is this the org url returned AFTER oauthing that includes the org_id? or is this the url that the customer gave at oauth time (e.g. login.salesforce.com or blah-blah.my.salesforce.com)?

we want to be sure this includes the OrgId or we use the orgid instead. multiple orgs can have the same login url

metadeploy/api/hash_url.py
```diff
@@ -0,0 +1,10 @@
+import hashlib
+
+
+def convert_org_url_to_key(url):
+    """
+    This isn't to securely hide the URL or anything; just to quickly make sure that we
+    can use it as a Redis key. That is, to make sure it doesn't contain any characters
+    Redis doesn't like.
+    """
+    return hashlib.sha1(url.encode("utf-8")).hexdigest()
```
cdcarter: is there a reason to actually hash it instead of just b64 encode it?

metadeploy/api/jobs.py
```diff
@@ -90,6 +90,7 @@ def mark_canceled(result):
         yield
     except StopRequested:
         result.status = result.Status.canceled
+        result.canceled_at = timezone.now()
```
cdcarter: hmm... technically this is when the job stopped itself, not when the cancel request was submitted... loose requirements. i think that's okay but we should add a help_text or something on the model so we all remember this.

metadeploy/api/models.py
```diff
@@ -445,21 +448,36 @@ def skip_tasks(self):
...
+        self._push_if_condition(results_has_changed, notify_post_task)
+
+    def push_if_has_stopped_running(self):
+        has_stopped_running = (
+            self.tracker.has_changed("status") and self.status != Job.Status.started
+        )
+        self._push_if_condition(has_stopped_running, notify_post_job)
+
     def save(self, *args, **kwargs):
+        is_new = bool(self.id)
```
cdcarter: prefer `self._state.adding` (it's not private, it's just _'d to not conflict...)

metadeploy/api/push.py
wlonk: It's the former, it comes from the SocialAuth's `extra_data` field.

metadeploy/api/hash_url.py
wlonk: b64 can still include `=`, which Redis doesn't like. That's what I tried at first :)

metadeploy/api/jobs.py
wlonk: Good point.

metadeploy/api/push.py
```diff
@@ -13,10 +13,14 @@
         JOB_COMPLETED
         JOB_FAILED
         JOB_CANCELED
+    org.:org_url
+        PREFLIGHT_CHANGED
+        JOB_CHANGED
```
jgerigmeyer: @wlonk Since these are specific to the structure of the `OrgSerializer`, I wonder if we can name the events more clearly? Maybe even combining them into one `ORG_CHANGED`, or `ORG_JOB_CHANGED` and `ORG_PREFLIGHT_CHANGED`.

wlonk: @jgerigmeyer No, I'm OK with this.

metadeploy/api/push.py
```diff
@@ -130,8 +129,8 @@
 
 
 async def notify_org_job_changed(job):
-    await _job_or_preflight_changed(job.organization_url, "JOB_CHANGED")
+    await _job_or_preflight_changed(job.organization_url, "ORG_CHANGED")
 
 
 async def notify_org_preflight_changed(preflight):
-    await _job_or_preflight_changed(preflight.organization_url, "PREFLIGHT_CHANGED")
+    await _job_or_preflight_changed(preflight.organization_url, "ORG_CHANGED")
```
wlonk: At this point, this can be one function, with "result" passed in generically.

PR #149, opened 2018-12-14
🐈 [hotfix-null-preflights-jobs] Hotfix error on missing preflights/jobs.

jgerigmeyer:
@wlonk This fixes an overlooked bug (introduced in https://github.com/SFDO-Tooling/metadeploy/commit/910c5d6f1e835db257cc71825f8bc32af20eb919) that broke fetching missing/`null` preflights and jobs.


jgerigmeyer:
(The dependency upgrades are superfluous -- just for giggles!)

PR #152, opened 2018-12-18
[http-logging] Log Heroku request ID

wlonk:
Imagine a cute animal.


wlonk:
@cdcarter another review would be appreciated, thanks!

PR #162, opened 2019-01-04
[integration-tests] 📦 Add basic integration tests that touch an SF org

wlonk:
sighthounds

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,4 @@
 [run]
 omit = 
+    metadeploy/tests/integration.py
     metadeploy/*/migrations/*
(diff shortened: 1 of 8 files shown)


wlonk:
@cdcarter It's not something you can run unless you have database access already, and in that case, that data is available to you. Is it just a matter of not making it _easy_?

cdcarter:
If we want a tool for extracting user credentials from the database in prod, then we'll need to put auditing and logging specifically for it in place.


...yes, we need that for any form of dyno access, but a named tool for getting sf prod credentials...raises...red flags... I think we'll have to work out operator acls internally here...

cdcarter:
so... merge it... we'll backlog any chanegs

PR #165, opened 2019-01-08
[signal-no-permissions] Show warning when current user lacks org permissions

wlonk:
maine coon

### DIFF
--- a/src/js/components/plans/detail.js
+++ b/src/js/components/plans/detail.js
@@ -20,17 +20,15 @@ import { shouldFetchVersion, getLoadingOrNotFound } from 'products/utils';
 import { startJob } from 'jobs/actions';
 
 import BodyContainer from 'components/bodyContainer';
-import CtaButton from 'components/plans/ctaButton';
+import CtaButton, { LoginBtn } from 'components/plans/ctaButton';
 import Header from 'components/plans/header';
 import Intro from 'components/plans/intro';
-import { ErrorIcon } from 'components/plans/jobResults';
-import JobResults from 'components/plans/jobResults';
+import JobResults, { ErrorIcon } from 'components/plans/jobResults';
 import PlanNotAllowed from 'components/products/notAllowed';
 import ProductNotFound from 'components/products/product404';
 import StepsTable from 'components/plans/stepsTable';
 import Toasts from 'components/plans/toasts';
 import UserInfo from 'components/plans/userInfo';
-import Login from 'components/header/login';
 
 import type { AppState } from 'app/reducer';
 import type { InitialProps } from 'components/utils';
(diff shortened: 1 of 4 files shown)


src/js/components/plans/detail.js
```diff
@@ -186,6 +189,30 @@ class PlanDetail extends React.Component<Props, State> {
       return <ProductNotFound />;
     }
     const selectedSteps = this.getSelectedSteps();
+    let cta = null;
+    if (user && !user.org_type) {
+      cta = (
+        <Login
+          id="user-info-login"
```
jgerigmeyer: This should be an `id` unique to this usage... maybe `org-not-allowed-login` or `missing-org-info-login` or something.

src/js/components/plans/detail.js
```diff
@@ -186,6 +189,30 @@ class PlanDetail extends React.Component<Props, State> {
...
+          label="log in with a different org"
```
jgerigmeyer: `Log in...`

src/js/components/plans/detail.js
```diff
@@ -186,6 +189,30 @@ class PlanDetail extends React.Component<Props, State> {
...
+          buttonVariant="base"
+        />
```
jgerigmeyer: I think this component should have basically the same settings as the `LoginBtn` component in `ctaButton.js`:

```js
  <Login
    id="org-not-allowed-login"
    buttonClassName="slds-size_full slds-p-vertical_xx-small"
    buttonVariant="brand"
    triggerClassName="slds-size_full"
    label="Log in with a different org"
    nubbinPosition="top"
  />
```

src/js/components/plans/detail.js
```diff
@@ -208,20 +235,29 @@ class PlanDetail extends React.Component<Props, State> {
...
+                        <span>
+                          Oops! It looks like you don&apos;t have permissions to
+                          run an installation on this org.
+                        </span>
+                      </div>
+                      <div>
+                        Please contact an Admin within your org or use the
+                        button below to log in with a different org.
+                      </div>
+                    </div>
```
jgerigmeyer: This can use the `ErrorIcon` component from `jobResults.js`, and generally follow the same pattern as the `ErrorsList` component in that file...

```js
<div>
  <ErrorIcon />
  <span className="slds-text-color_error">
    Oops! It looks...
  </span>
  <div>
    Please contact...
  </div>
</div>
```

src/js/components/plans/detail.js
wlonk: Oop, that was copypasta! I'll fix.

src/js/components/plans/detail.js
wlonk: you and your capital letters

src/js/components/plans/detail.js
jgerigmeyer: Bonus points if you want to just export the component from `ctaButton`, making its `id` customizable (probably with a default of "plan-detail-login").

src/js/components/plans/detail.js
```diff
@@ -23,12 +23,14 @@ import BodyContainer from 'components/bodyContainer';
...
 import Header from 'components/plans/header';
 import Intro from 'components/plans/intro';
+import { ErrorIcon } from 'components/plans/jobResults';
 import JobResults from 'components/plans/jobResults';
 import PlanNotAllowed from 'components/products/notAllowed';
 import ProductNotFound from 'components/products/product404';
 import StepsTable from 'components/plans/stepsTable';
 import Toasts from 'components/plans/toasts';
 import UserInfo from 'components/plans/userInfo';
+import Login from 'components/header/login';
```
jgerigmeyer: Obviously minor, but convention is to alphabetize imports, and put partial imports below default imports (so `{ ErrorIcon }` would come last).

wlonk: Oh, you don't have an `isort` precommit hook on the JS side, right. I kinda operated on the "tools will fix this" assumption!

PR #166, opened 2019-01-09
[no-unstyled-auth-errors] Render all socialaccount templates as nice error pages

wlonk:
Not sure how to test all plausible errors, so I'd appreciate some more hand-testing.

See https://trello.com/c/dgbaoDb4/77-2-as-an-admin-with-a-newly-created-dev-org-i-want-a-helpful-styled-error-message-if-i-try-to-log-in-too-quickly

bongo


wlonk:
I thought we didn't have any non-socialaccount accounts, so I didn't include that. I also guess I didn't understand how we were currently rendering error messages; was that relying on the path for the frontend to guess "if I'm rendering this it's an error"?

I agree with your second paragraph; I think that's more correct.

jgerigmeyer:
@wlonk Yes, your understanding of the way the JS works now is correct. And I think you're right about the socialaccounts, but allauth adds other URLs that I don't think we want to expose unstyled.

I've got my proposal working here locally, so I'll push that shortly.

wlonk:
👍 I like it.

PR #167, opened 2019-01-09
[continue-job] Display failed state on job more clearly

wlonk:
See https://trello.com/c/kFUYhdz0/92-as-an-admin-with-a-failed-installation-i-want-the-failed-state-to-be-more-obvious

otter


wlonk:
@jgerigmeyer I'm not sure how to write good tests for this, so I'd appreciate input on that; hand testing was good though.

PR #169, opened 2019-01-10
[i18n] Begin the internationalization

wlonk:
See https://trello.com/c/kET0lQTK/12-3-as-salesforce-staff-i-want-to-ensure-proper-creation-and-use-of-i11n-able-strings-for-all-user-facing-copy

nothing more international than a wombat


jgerigmeyer:
@wlonk I'm not sure when this is considered "done", at least for the back-end? Do we also want .po files?

wlonk:
I'm unsure! I'll see if people think checking in the .po files is a good practice.

src/js/index.js
```diff
@@ -39,6 +39,30 @@ import JobDetail from 'components/jobs/detail';
 import PlanDetail from 'components/plans/detail';
 import ProductsList from 'components/products/list';
 import { ProductDetail, VersionDetail } from 'components/products/detail';
+import i18n from 'i18next';
+import { reactI18nextModule } from 'react-i18next';
+
+i18n
+  .use(reactI18nextModule) // passes i18n down to react-i18next
```
jgerigmeyer: I think we may also want to auto-detect here (https://react.i18next.com/guides/step-by-step-guide#c-auto-detect-the-user-language), and possibly also use lazy-loading (with a fallback to en):

- https://react.i18next.com/guides/step-by-step-guide#2-lazy-loading-translations
- https://react.i18next.com/guides/step-by-step-guide#set-the-global-wait-option

src/js/index.js
```diff
@@ -39,6 +39,30 @@ import JobDetail from 'components/jobs/detail';
...
+  .init({
+    resources: {
+      en: {
+        translation: {
+          // We just put key: value mappings for English here. Like:
+          // "Log In": "Go Forth"
+          // Anything missing will fall back to the key.
+          // What I don't know is how to abstract this out and populate its
+          // keys, à la makemessages.
+        },
```
jgerigmeyer: These should definitely get pulled out into separate files (e.g. https://react.i18next.com/guides/step-by-step-guide#c-auto-detect-the-user-language), and probably also auto-generated initially (https://react.i18next.com/guides/extracting-translations).

src/js/components/header/login.js
```diff
@@ -145,4 +149,6 @@ class Login extends React.Component<Props, { modalOpen: boolean }> {
   }
 }
 
+const Login = withI18n()(BaseLogin);
```
jgerigmeyer: It'll be interesting to see how well this works for components that are already wrapped with redux `connect` HOCs... We may want to use the `lodash.compose` pattern shown here: https://reactjs.org/docs/higher-order-components.html#convention-maximizing-composability

src/js/components/header/login.js
```diff
@@ -116,11 +119,12 @@ class Login extends React.Component<Props, { modalOpen: boolean }> {
       redirectParams,
     } = this.props;
     const { modalOpen } = this.state;
+    const { t } = this.props;
```
jgerigmeyer: This should be combined with the list above on lines 110-120.

src/js/components/header/login.js
```diff
@@ -116,11 +119,12 @@ class Login extends React.Component<Props, { modalOpen: boolean }> {
...
     return (
       <>
         <Dropdown
           id={id}
-          label={label}
+          label={t(this.props.label)}
```
jgerigmeyer:
```suggestion
          label={t(label)}
```

locale/en_US/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,84 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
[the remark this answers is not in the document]
jgerigmeyer: Hm, I guess I'm not sure why this is committed in that case.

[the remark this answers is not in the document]
wlonk: Because I'm still trying to figure out best practices around this; I'm not sure if the i18n side of Django will complain if we try to access translations for a language and don't have a translation file for it.

CONTRIBUTING.rst
```diff
@@ -186,3 +186,31 @@ need to generate scratch orgs to test against.
...
+   $ python manage.py compilemessages
+
+These commands require the `GNU gettext toolset`_ (``brew install gettext``).
+
+For the front-end, translation JSON files are served from
+``locales/<language>/`` directories, and the `user language is auto-detected at
+runtime`_.
+
+During development, strings are parsed automatically from the JS, and an English
+translation file is auto-generated to ``locales/dev/en/translation.json`` on
```
jgerigmeyer: maybe move into its own dir?

__mocks__/i18next.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-env jest */
+
+'use strict';
+
+const i18n = jest.genMockFromModule('i18next');
+
+i18n.t = jest.fn(str => str);
```
jgerigmeyer: maybe do the real translation in tests?

src/js/components/404.js
```diff
@@ -20,14 +21,14 @@ export const EmptyIllustration = ({ message }: { message: React.Node }) => (
 );
 
 const FourOhFour = ({ message }: { message?: React.Node }) => (
-  <DocumentTitle title="404 | MetaDeploy">
+  <DocumentTitle title={`${t('404')} | ${t('MetaDeploy')}`}>
```
wlonk: Do these need to be translated?

src/js/components/footer.js
```diff
@@ -25,7 +26,7 @@ const Footer = (props: { logoSrc: string }) => (
       className="footer-item
         slds-grid"
     >
-      <p>Copyright 2000–2018 Salesforce.org. All rights reserved.</p>
+      <p>{t('Copyright 2000–2018 Salesforce.org. All rights reserved.')}</p>
```
jgerigmeyer: update to 2019

src/js/components/header/login.js
```diff
@@ -2,17 +2,16 @@
...
 import { logError } from 'utils/logging';
-
 import CustomDomainModal from 'components/header/customDomainModal';
-
 import type { UrlParams } from 'utils/api';
 
 type Props = {
   id: string,
-  label: string | React.Node,
+  label?: string | React.Node,
```
wlonk: Why is label now optional, or is it not really? Is this something about `t` and language loading? Maybe add comment.

wlonk: Maybe we could make translations load synchronously, and that'll make things cleaner elsewhere?

src/js/components/offlineAlert.js
```diff
@@ -12,10 +13,10 @@ const OfflineAlert = () => (
...
     <Alert
       labels={{
-        heading:
-          'You are in offline mode. We are trying to reconnect, ' +
-          'but you may need to',
-        headingLink: 'reload the page.',
+        heading: t(
+          'You are in offline mode. We are trying to reconnect, but you may need to',
+        ),
+        headingLink: t('reload the page.'),
```
wlonk: There's no guarantee that in a translation, these will be two adjacent parts of a sentence in a fixed order. But this component ain't ours, so.

src/js/components/plans/stepsTable/installDataCell.js
```diff
@@ -107,11 +108,11 @@ const JobCell = (props: DataCellProps): React.Node => {
           >
             <Spinner size="small" />
           </span>
-          Installing...
+          {t('Installing...')}
```
wlonk: `&hellip;` or `…` instead of `...`.

test/js/components/plans/jobResults.test.js
```diff
@@ -120,7 +120,7 @@ describe('<JobResults />', () => {
 
       expect(
         getByText(
-          'Pre-install validation encountered 4 errors and 3 warnings.',
+          'Pre-install validation encountered errorMsg and warningMsg.',
```
wlonk: Give these sensible translation defaults.

jgerigmeyer:
@davisagli I don't think you need to review this PR in detail (it's a lot of repetitive changes), but maybe worth reading the new docs to see how it's setup to work? https://github.com/SFDO-Tooling/metadeploy/pull/169/files#diff-ea1f65a8eaac0cec7ae151dfdbbbc587R190

I'll also note that this does _not_ add translation for front-end user-facing strings controlled by third-parties (notably design-system-react). In some cases we could remedy that by passing in our own strings to components instead of using their defaults, but in other cases it would require a completely different approach.

PR #177, opened 2019-01-16
[progress-indicator] Add plan page progress indicator.

wlonk:
cotton


jgerigmeyer:
@mirisuzanne 

1. As far as I can tell, SLDS doesn't have a pattern for labels other than tooltips... and it looks pretty busy with the tooltips all open.
2. Yeah, that's because the entire thing has a `max-width: 70%` set on it, so the tooltip is being compressed. Is there a good way to fix that with CSS?
3. Hm, do you have an alternate proposal for when a preflight expires? Should the "Run pre-install validation" step become unchecked? Or the last step just become unchecked instead of errored?

mirisuzanne:
@jgerigmeyer 
1. ok
2. if they have a non-center tooltip option, giving the final one a right-edge-arrow, and letting it flow left might solve the problem.
3. I think unchecking would be fine: back to whichever state it was in before the initial preflight. Or replacing the red cross-out with a yellow/orange warning icon?

jgerigmeyer:
@mirisuzanne Okay, (3) is fixed, and I'm submitted an issue (and PR fix) to design-system-react for (2): https://github.com/salesforce/design-system-react/issues/1720

PR #179, opened 2019-01-17
🛢📦[stepspec] Add StepSpec members to Step

cdcarter:
Moving toward the "flowprovements" branch of cci, we're gonna need more fields on the Step model to match the StepSpec api.

img_0122


metadeploy/api/models.py
```diff
@@ -389,16 +388,25 @@ class Step(HashIdMixin, models.Model):
...
     plan = models.ForeignKey(Plan, on_delete=models.PROTECT)
-    name = models.CharField(max_length=1024)
+    name = models.CharField(max_length=1024, help_text="Customer facing label")
     description = models.TextField(blank=True)
     is_required = models.BooleanField(default=True)
     is_recommended = models.BooleanField(default=True)
     kind = models.CharField(choices=Kind, default=Kind.metadata, max_length=64)
-    order_key = models.PositiveIntegerField(default=0)
-    task_name = models.CharField(max_length=64)
+    task_name = models.CharField(max_length=64, help_text="CCI Config task_name")
```
davisagli: I think we'll need the `path`, not just the `task_name`, in order to successfully run tasks whose options have ^^ references to other tasks' return values. (probably not an immediate blocker)

metadeploy/api/models.py
```diff
@@ -389,16 +388,25 @@ class Step(HashIdMixin, models.Model):
     )
 
     plan = models.ForeignKey(Plan, on_delete=models.PROTECT)
```
davisagli: Can we switch to `steps` as the related_name? step_set makes it sound like the order doesn't matter.

metadeploy/api/models.py
```diff
@@ -389,16 +388,25 @@ class Step(HashIdMixin, models.Model):
...
+        max_length=2048, help_text="dotted module path to BaseTask implementation"
+    )
+    task_config = JSONField(default=dict, blank=True)
 
     class Meta:
-        ordering = ("order_key", "name")
+        ordering = (
+            "step_num",
+            "task_name",
+        )  # TODO: does postgres sort step_num the way I expect
```
davisagli: https://stackoverflow.com/questions/31089961/ordering-dot-delimited-numeric-sequences-e-g-version-numbers

Why task_name here, btw?

metadeploy/api/cci_configs.py
```diff
@@ -28,5 +29,24 @@ def __init__(self, *args, repo_root=None, plan=None, **kwargs):  # pragma: nocov
         super().__init__(*args, repo_info=repo_info, **kwargs)
 
 
+class MetaDeployCallback(FlowCallback):
+    pass
+
+
 class MetaDeployCCI(BaseCumulusCI):
     project_config_class = MetadeployProjectConfig
+    callback_class = FlowCallback
```
davisagli: You defined MetaDeployCallback but aren't using it

metadeploy/api/models.py
cdcarter: i like the suggestion of cast to CIDR :P . long as we're under four levels of nesting ;)

davisagli: yeah string_to_array seems safer. Looks like Django supports using query expressions for `ordering`

metadeploy/api/models.py
cdcarter: done in latest push

metadeploy/api/models.py
cdcarter: that was a little bit more fun than i wanted it to be, and it's not fully done yet, but take a look. This will likely (certainly) be something we can make useful and reusable considering how many looseversion and dotted.name style stuff we're using here.

metadeploy/api/cci_configs.py
```diff
@@ -30,3 +35,27 @@ def __init__(self, *args, repo_root=None, plan=None, **kwargs):  # pragma: nocov
...
     project_config_class = MetadeployProjectConfig
+
+    def get_flow_from_plan(
+        self, plan: Plan, ctx: WorkableModel, skip: List[str] = None
+    ):
+
+        steps = [
+            step.to_spec(skip=True) if step.path in skip else step.to_spec(skip=False)
+            for step in plan.steps
+        ]
```
cdcarter: @davisagli is this too unseemly for you?

metadeploy/api/cci_configs.py
```diff
@@ -30,3 +35,27 @@ def __init__(self, *args, repo_root=None, plan=None, **kwargs):  # pragma: nocov
...
+
+        # TODO: either use the dynamic stuff i put into baseruntime, or scrap it.
```
cdcarter: yea, I am going to remove the get_callback_class stuff in BaseRuntime. The default #get_flow(flow_config) method will still be able to get a declared callback_cls from the class variables, there just won't be anotehr level of redirection.

metadeploy/api/flows.py
```diff
@@ -15,36 +15,35 @@ class StopFlowException(Exception):
...
-        return super().__init__(*args, **kwargs)
+class MetaDeployCallback(FlowCallback):
+    def __init__(self, ctx):
+        self.context = ctx  # will be either a preflight or a job...
+        # TODO: collate results, probably requires a change in how
+        # FlowCoordinator calls its Callbacks....
 
-    def _get_step_id(self, task_name):
+
+class BasicFlow(MetaDeployCallback):
```
cdcarter: .... Merge this and MetaDeployFLowCallback....

metadeploy/api/flows.py
```diff
@@ -15,36 +15,35 @@ class StopFlowException(Exception):
...
             raise StopFlowException("Job canceled.")
 
     def _flow_canceled(self):
-        return cache.get(REDIS_JOB_CANCEL_KEY.format(id=self.result.id))
+        return cache.get(REDIS_JOB_CANCEL_KEY.format(id=self.context.id))
 
 
 class JobFlow(BasicFlow):
     def _init_logger(self):
+        # TODO FIXME!!!
```
cdcarter: specifically, you can't hack the logger in the override of the callbacks now. that's...something elses problem. TODO figure out what...

metadeploy/api/flows.py
```diff
@@ -15,36 +15,35 @@ class StopFlowException(Exception):
...
+    def pre_task(self, step):
         """
         Before each task, we should check if we've been told to abandon this job.
         """
         if self._flow_canceled():
             raise StopFlowException("Job canceled.")
 
     def _flow_canceled(self):
-        return cache.get(REDIS_JOB_CANCEL_KEY.format(id=self.result.id))
+        return cache.get(REDIS_JOB_CANCEL_KEY.format(id=self.context.id))
```
cdcarter:
```
def post_task(self, task, result):
   self.results.append(result)
```

metadeploy/api/cci_configs.py
davisagli: @cdcarter Can I talk you into `[step.to_spec(skip=step.path in skip) for step in plan.steps]` (same result) or `[step.to_spec() for step in plan.steps if step.path not in skip]` (leaves out the skipped steps altogether but they wouldn't be executed anyway)

cdcarter: great minds.

i went with the former, which will be more in line with the existing frontend's expectation IMO

metadeploy/adminapi/api.py
```diff
@@ -170,26 +170,26 @@ class ProductSlugViewSet(AdminAPIViewSet):
 class PlanStepSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Step
-        exclude = ("id", "plan", "order_key")
+        exclude = ("id", "plan")
```
wlonk: I have a medium-strong preference for explicitly enumerating the included fields with `fields`, rather than using `exclude`, but not enough to say "redo this".

metadeploy/api/flows.py
```diff
@@ -54,30 +51,31 @@ def _init_logger(self):
...
-            self.result.log = obscure_salesforce_log(self.string_buffer.getvalue())
-            self.result.save()
-        return super()._post_task(task)
+            self.context.results[step_id] = [{"status": OK}]
+            self.context.log = obscure_salesforce_log(self.string_buffer.getvalue())
+            self.context.save()
+        return super().post_task(step, result)
 
     def _post_task_exception(self, task, exception):
+        # TODO FIXME!!!
```
wlonk: What needs fixing here?

metadeploy/api/flows.py
```diff
@@ -54,30 +51,31 @@ def _init_logger(self):
         self.logger = logger
         return self.logger
 
-    def _post_flow(self):
+    def post_flow(self):
```
davisagli: Needs to take coordinator as a 2nd arg

metadeploy/api/flows.py
```diff
@@ -54,30 +51,31 @@ def _init_logger(self):
...
-        return super()._post_task_exception(task, exception)
+            self.context.log = obscure_salesforce_log(self.string_buffer.getvalue())
+            self.context.save()
+        # return super()._post_task_exception(task, exception)
 
 
-class PreflightFlow(BasicFlow):
-    def _post_flow(self):
+class PreflightFlowCallback(BasicFlowCallback):
+    def post_flow(self):
```
davisagli: Needs to take coordinator as a 2nd arg

metadeploy/api/flows.py
cdcarter: the new FlowCallback api expects  handle task failure in post_task, instead of its own post_task_exception method

metadeploy/adminapi/api.py
cdcarter: since its in the adminapi I think its alright.

metadeploy/api/flows.py
```diff
@@ -54,30 +51,31 @@ def _init_logger(self):
...
+        step_id = self._get_step_id(step.path)
         if step_id:
-            self.result.results[step_id] = [{"status": OK}]
-            self.result.log = obscure_salesforce_log(self.string_buffer.getvalue())
-            self.result.save()
-        return super()._post_task(task)
+            self.context.results[step_id] = [{"status": OK}]
+            self.context.log = obscure_salesforce_log(self.string_buffer.getvalue())
+            self.context.save()
+        return super().post_task(step, result)
```
davisagli: We know there isn't a super method that does anything, so remove this?

metadeploy/api/models.py
```diff
@@ -388,17 +399,27 @@ class Step(HashIdMixin, models.Model):
...
-    task_name = models.CharField(max_length=64)
+    path = models.CharField(
+        max_length=2048, help_text="dotted path e.g. flow1.flow2.task_name"
+    )
+    step_num = models.CharField(
+        max_length=64, help_text="dotted step number for CCI task"
+    )
+    task_class = models.CharField(
+        max_length=2048, help_text="dotted module path to BaseTask implementation"
+    )
```
davisagli: @cdcarter I don't think we need this as a separate field. `class_path` is already part of the `task_config`

config/settings/base.py
```diff
@@ -271,6 +271,7 @@ def env(name, default=NoDefaultValue, type_=str):
 }
 ACCOUNT_EMAIL_REQUIRED = True
 ACCOUNT_UNIQUE_EMAIL = False
+ACCOUNT_EMAIL_VERIFICATION = "none"
```
cdcarter: thankyou

metadeploy/api/flows.py
```diff
@@ -51,31 +48,24 @@ def _init_logger(self):
         self.logger = logger
         return self.logger
```
cdcarter: remove return

cdcarter: on my local branch I actually left this in _init_logger and just called that from pre_flow....

cdcarter: whatever works

PR #184, opened 2019-01-22
[launch-preflight] Trigger preflight on page load

wlonk:
But tests are failing.


src/js/components/plans/ctaButton.js
```diff
@@ -97,6 +97,13 @@ class CtaButton extends React.Component<Props, { modalOpen: boolean }> {
     this.state = { modalOpen: false };
   }
 
+  componentDidMount() {
+    const { plan, doStartPreflight, user, preflight } = this.props;
+    if (user && user.valid_token_for && !(preflight && preflight.is_valid)) {
+      doStartPreflight(plan.id);
+    }
+  }
```
jgerigmeyer: @wlonk My understanding from the Trello card is that we only want this to happen after clicking "Log In to Start Pre-Install Validation" button, not on every page-load. I think we talked about doing this with a query-string passed through the login flow?

wlonk:
@jgerigmeyer ah, that sounds right. How do you propose the query string to work?

jgerigmeyer:
@wlonk Tests will still fail here (I haven't touched them yet), but in the meantime I'd appreciate some eyes on the code (and functionality too).

src/js/components/plans/ctaButton.js
```diff
@@ -91,10 +96,65 @@ export const ActionBtn = ({
...
+      // `preflight === null`: no prior preflight exists
+      // `preflight === undefined`: still fetching prior preflights
+      // If we don't know about past preflights yet, wait until we do...
+      if (preflight === undefined) {
+        this.setState({ startPreflight: true });
+      } else {
+        this.autoStartPreflight();
+      }
+      // Remove query-string from URL
+      history.replace({ search: '' });
```
wlonk: Does this generate a back-button entry? Or would that only be on `pushState`?

src/js/components/plans/ctaButton.js
```diff
@@ -91,10 +96,65 @@ export const ActionBtn = ({
...
+    const preflightChanged = preflight !== prevProps.preflight;
+    if (startPreflight && preflightChanged && preflight !== undefined) {
+      this.autoStartPreflight();
+    }
+  }
+
+  autoStartPreflight(): void {
+    const { plan, doStartPreflight } = this.props;
+    this.setState({ startPreflight: false });
+    if (this.canStartPreflight()) {
```
wlonk: Ugh so much cleaner than what I wrote, a small well-named function that describes what it means, thank you!

src/js/components/plans/ctaButton.js
jgerigmeyer: This replaces the current entry. `history.push()` would create a new entry.

PR #189, opened 2019-01-24
[hotfix-login-dropdowns] Prevent login dropdowns from disappearing outside scrollable area.

jgerigmeyer:
oops


src/js/components/plans/ctaButton.js
```diff
@@ -118,7 +119,7 @@ class CtaButton extends React.Component<
         this.autoStartPreflight();
       }
       // Remove query-string from URL
-      history.replace({ search: removeUrlParam(AUTO_START_PREFLIGHT) });
+      history.push({ search: removeUrlParam(AUTO_START_PREFLIGHT) });
```
jgerigmeyer: This isn't directly related to this branch/PR, but I noticed that a `replace` which only changes the search string apparently does not persist through a page-reload... such that on every reload the `?start_preflight=true` would re-appear. This change adds a history entry, but I think that's preferable to unintentionally running another preflight on reload.

metadeploy/api/management/commands/populate_data.py
```diff
@@ -46,7 +46,8 @@ def create_version(self, product, label="0.3.1", **kwargs):
 
     def create_plan(self, version, title="Full Install", tier="primary", **kwargs):
         combined_kwargs = {
-            "preflight_flow_name": "static_preflight",
+            # "preflight_flow_name": "static_preflight",
+            "preflight_flow_name": "",
```
jgerigmeyer: Not directly related to this PR, but allows development to continue while preflights are being fixed in CCI.

PR #191, opened 2019-01-25
📦 switch to a middleware for ip ranges

cdcarter:
validated manually locally. I'd like to make sure the `django-ipware` defaults work on Heroku. it's a very lightweight dependency (with no deps of its own) that has well reasoned logic for detecting the client IP. 

got rid of the scary permissions factory method. all admin prefixed routes get protected by this instead of different mechanisms protecting api vs admin interface.

we should definitely backport this to the template.

img_0207

### DIFF
--- a/metadeploy/admin_middleware.py
+++ b/metadeploy/admin_middleware.py
@@ -1,0 +1,28 @@
+from ipaddress import IPv4Address
+
+from django.conf import settings
+from django.core.exceptions import SuspiciousOperation
+from ipware import get_client_ip
+
+
+class AdminRestrictMiddleware:
+    """
+    A middleware that restricts all access to the admin prefix to allowed IPs.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        self.ip_ranges = settings.ADMIN_API_ALLOWED_SUBNETS
+
+    def __call__(self, request):
+        if request.path.startswith(f"/{settings.ADMIN_AREA_PREFIX}"):
+            self._validate_ip(request)
+
+        return self.get_response(request)
+
+    def _validate_ip(self, request):
+        ip_str, _ = get_client_ip(request)
+        ip_addr = IPv4Address(ip_str)
+
+        if not any(ip_addr in subnet for subnet in self.ip_ranges):
+            raise SuspiciousOperation(f"Disallowed IP address: {ip_addr}")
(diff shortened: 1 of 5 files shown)


metadeploy/admin_middleware.py
```diff
@@ -0,0 +1,28 @@
...
+        self.ip_ranges = settings.ADMIN_API_ALLOWED_SUBNETS
+
+    def __call__(self, request):
+        if request.path.startswith(f"/{settings.ADMIN_AREA_PREFIX}"):
+            self._validate_ip(request)
+
+        return self.get_response(request)
+
+    def _validate_ip(self, request):
+        ip_str, _ = get_client_ip(request)
```
davisagli: Don't we need to specify `proxy_count` here?

cdcarter:
Oops, I'll fix the missing coverage.

Yes, we want to block the old school django admin app, and at that point doing it centrally by path was easy!

metadeploy/__init__.py
```diff
@@ -0,0 +1,11 @@
+from functools import partial
+
+from ipware import get_client_ip as get_ip_original
+
+get_client_ip = partial(get_ip_original, proxy_order="right_most")
```
cdcarter: @wlonk normally I would put this in a utils.py but since we don't have one yet, I just put it here. do you have a better place for it, knowing our pkg structure better than I do?

metadeploy/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+get_client_ip.__name__ = "get_client_ip"
+get_client_ip.__doc__ = (
+    "Retrieve the client's IP from a request object.\n\n"
+    "using the 'right most' proxy ordering in order to get the IP that originated "
+    "the request to the Heroku LB."
+)
```
davisagli: I'm not convinced this is better than actually defining a new function with `def`, but whatever.

metadeploy/__init__.py
davisagli: I think adding a utils.py would be fine.

wlonk: Here's my hierarchy of rathering:

well-named module (even if the name is a duplicate of the function it contains) > `utils.py` > meaningful code in `__init__.py`

metadeploy/__init__.py
wlonk: I'm with David here.

PR #192, opened 2019-01-25
[hotfix-long-product-descriptions] Only show first element in product-list description (truncated).

jgerigmeyer:
ostrich


jgerigmeyer:
Before:

screenshot 2019-01-25 18 00 03

With this PR:

screenshot 2019-01-25 18 00 39

src/sass/patterns/_icons.scss
```diff
@@ -1,7 +1,7 @@
-// Products
-// ========
+// Icons
+// =====
 
-.slds-icon-standard-account {
+[class*='slds-icon'] {
```
jgerigmeyer: Hm, I'm nervous about this change. That'll target a lot of internal SLDS-applied classes.

mirisuzanne: yeah – really I'm worried about hooking into `slds` classes at all – sort of breaks the point of a design system to have a class do something unexpected like this. Is there any way for us to apply our own hooks on the icon? Or (even better) apply the style attr to the icon itself?

mirisuzanne:
@jgerigmeyer I figured `class*=slds-icon` might target multiple things, but it's also nested inside `style*=--custom-color` which is a pretty narrow set. And this makes the pattern more flexible for any icons we want to custom-color. ¯\_(ツ)_/¯

PR #198, opened 2019-01-29
Added configuration for objects in the Django Admin site

jlantz:


jlantz:
Yeah, this branch actually helps us in that regard.  For example, the select fields for Plan or even worse the Steps field on a Job would eventually crater under data volume.  This PR replaces those fields which are either ForeignKey or ManyToMany relationships to models likely to have many rows of data with the new autocomplete2 widget designed as a solution to that problem.

I also made sure not to put anything in list_filter that would have a lot of rows and tried to properly use list_select_related where the list_display fields have references to related models.

PR #201, opened 2019-01-30
use admin area env var

cdcarter:
respect the `DJANGO_ADMIN_URL`


config/settings/base.py
```diff
@@ -28,6 +28,12 @@ def ipv4_networks(val: str) -> List[IPv4Network]:
     return [IPv4Network(s.strip()) for s in val.split(",")]
 
 
+def url_prefix(val: str) -> str:
+    if val[-1] != "/":
+        val += "/"
+    return val
```
davisagli: return val.rstrip("/") + "/"

PR #203, opened 2019-01-31
[short_description] 🛢 Add short_description to Products

wlonk:
See https://trello.com/c/fRdw5Niw/103-add-product-shortdescription

doggo


jgerigmeyer:
@mirisuzanne @cdcarter @davisagli For the moment this is using `product.description` (truncated) as a fallback if `product.short_description` is missing, and it is _not_ truncating the `short_description`. It is also only displaying the `short_description` on the products-list page, not product-detail. If preferred, we could truncate both in the list, and/or show both on the product-detail page.

screenshot 2019-01-31 16 40 36

davisagli:
@jgerigmeyer current approach sounds optimal to me.

PR #207, opened 2019-02-04
[model-translation] 📦 🛢 Add model field translations

wlonk:
kitten


metadeploy/adminapi/api.py
```diff
@@ -137,13 +137,19 @@ class ProductSlugViewSet(AdminAPIViewSet):
 
 
 class PlanStepSerializer(serializers.ModelSerializer):
+    name = serializers.CharField(required=False)
```
davisagli: name should be required

metadeploy/adminapi/api.py
```diff
@@ -137,13 +137,19 @@ class ProductSlugViewSet(AdminAPIViewSet):
...
     class Meta:
         model = models.Step
         exclude = ("id", "plan")
 
 
 class PlanSerializer(AdminAPISerializer):
     steps = PlanStepSerializer(many=True, required=False)
+    post_install_message = serializers.CharField(required=False)
+    preflight_message = serializers.CharField(required=False)
+    title = serializers.CharField(required=False)
```
davisagli: title should be required. For the others, is required=False okay even though the fields don't have null=True?

davisagli: Do we need to be explicit about the serializers for Product too? There isn't full functional test coverage of the admin API yet.

wlonk: @davisagli I'm not sure, because I don't know where and how Product serializers might be used. Can you tell me more?

davisagli: By the ProductViewSet at /admin/rest/products

wlonk: Sorry, I meant that in a slightly different way; I don't know if the intent is to populate those fields or read those fields (`title`, `short_description`, `description`, `click_through_agreement`) via the admin API. I'll assume it is, and just add them to the thing explicitly, but I think we should have more admin API test coverage.

davisagli: The intent of the admin API is to provide superusers with read and write access to all the fields. I agree about test coverage and am sorry it's not there yet. In practice we aren't using the API to populate products yet, so it's not high priority.

PR #210, opened 2019-02-07
Fix running preflight flows

davisagli:
This aims to get the existing preflight flows that were set up for UX testing working again.


wlonk:
`metadeploy.api.tests.flows.TestPreflightFlow.test_post_flow` is still failing.

davisagli:
@jgerigmeyer maybe this new commit helps?

davisagli:
Hmm, the test that's failing on CircleCI is passing for me locally. Which doesn't make sense to me because I also have a bogus GITHUB_TOKEN locally.

PR #211, opened 2019-02-08
[plan-text-restructure] 🛢 Add project template model

wlonk:
I suspect there are choices we'll want to make on all the serializers to better paper over this, but I don't think we have to automatically generate any ProjectTemplate instances, because that all happens through the AdminAPI, which should rather be explicit.

kittens


metadeploy/api/models.py
```diff
@@ -400,18 +415,23 @@ class Plan(HashIdMixin, SlugMixin, AllowedListAccessMixin, TranslatableModel):
...
-    def preflight_message_markdown(self):
-        return self.get_translation("en-us").preflight_message_markdown
+    def preflight_message_additional_markdown(self):
+        return self.get_translation("en-us").preflight_message_additional_markdown
 
     @property
-    def post_install_message_markdown(self):
-        return self.get_translation("en-us").post_install_message_markdown
+    def post_install_message_additional_markdown(self):
+        return self.get_translation("en-us").post_install_message_additional_markdown
```
jgerigmeyer: I don't see where this is being used; should it be added to the serializer?

metadeploy/api/serializers.py
```diff
@@ -140,6 +145,7 @@ class Meta:
             "title",
             "version",
             "preflight_message",
+            "preflight_message_additional",
```
jgerigmeyer: I assume we should also display this on the front-end?

metadeploy/api/migrations/0051_copy_plan_to_plan_template.py
```diff
@@ -0,0 +1,26 @@
...
+def forwards(apps, schema_editor):
+    Plan = apps.get_model("api", "Plan")
+    PlanTemplate = apps.get_model("api", "PlanTemplate")
+    for plan in Plan.objects.all():
+        plan_template = PlanTemplate.objects.create(
+            preflight_message=plan.preflight_message,
+            post_install_message=plan.post_install_message,
+        )
+        plan.plan_template = plan_template
+        plan.save()
```
davisagli: I'm thinking a better strategy for the migration may be to rename preflight_message to preflight_message_additional, rename post_install_message to post_install_message_additional, and not automatically create any plan templates.

metadeploy/api/models.py
```diff
@@ -360,6 +360,21 @@ def additional_plans(self):
         return self.plan_set.filter(tier=Plan.Tier.additional).order_by("id")
 
 
+class PlanTemplate(TranslatableModel):
```
davisagli: PlanTemplate should have a name (so that we can put that as an identifier in the configuration for the code that publishes a product's installer). The name probably has to be nullable and blankable for the sake of migration.

metadeploy/api/models.py
```diff
@@ -400,18 +415,23 @@ class Plan(HashIdMixin, SlugMixin, AllowedListAccessMixin, TranslatableModel):
...
+    def preflight_message_additional_markdown(self):
+        return self.get_translation("en-us").preflight_message_additional_markdown
 
     @property
-    def post_install_message_markdown(self):
-        return self.get_translation("en-us").post_install_message_markdown
+    def post_install_message_additional_markdown(self):
+        return self.get_translation("en-us").post_install_message_additional_markdown
 
+    plan_template = models.ForeignKey(PlanTemplate, on_delete=models.PROTECT)
```
davisagli: This should have null=True and blank=True, no?

wlonk: Should it? I thought that any Plan would always have a PlanTemplate.

metadeploy/api/models.py
wlonk: Or an auto-generated default, either way.

metadeploy/api/migrations/0051_copy_plan_to_plan_template.py
wlonk: So, that implies:

 - We start with no templates, and all content in the `_additional` fields.
 - The `plan_template` FK is nullable, and no Plans have a value there initially.
 - As you move things into templates, you can manually take data out of the `_additional` fields and put it in the template, and wire up the connection.
 - The frontend will display the template text and the `_additional` text concatenated, so this initial state won't be an issue.

Is that about right?

metadeploy/api/models.py
davisagli: I don't see a huge downside either way. But it doesn't seem like it _has_ to have one unless it wants to share text with other plans.

We can leave this as is.

metadeploy/api/serializers.py
wlonk: @jgerigmeyer Hold on for a moment, we might provide you with the already-concatenated value to display on the frontend, requiring no changes on your end.

metadeploy/api/migrations/0051_copy_plan_to_plan_template.py
davisagli: That's what I was picturing, yes.

metadeploy/api/models.py
wlonk: It is in the `JobSerializer`.

wlonk:
@davisagli and @jgerigmeyer I incorporated feedback and I think that the frontend needs to do nothing to adapt now, and the backend is stronger for it.

PR #219, opened 2019-02-12
[track-license-requirements] 🛢 Track the clickthrough agreement text

wlonk:
Save the original in a db-space-minimizing way, for later auditing.

See https://trello.com/c/V5Ff3oSg/105-as-salesforce-staff-i-want-revision-tracking-for-product-licenserequirements

kitten


metadeploy/api/models.py
```diff
@@ -529,6 +536,9 @@ class Job(HashIdMixin, models.Model):
     )
     exception = models.TextField(null=True)
     log = models.TextField(blank=True)
+    click_through_agreement = models.ForeignKey(
+        ClickthroughText, on_delete=models.PROTECT, null=True
+    )
```
davisagli: For consistency either the model should be called ClickThroughText or the field should be called clickthrough_text

metadeploy/api/models.py
```diff
@@ -502,6 +502,13 @@ def __str__(self):
         return f"Step {self.name} of {self.plan.title} ({self.step_num})"
 
 
+class ClickthroughText(models.Model):
+    text = models.TextField()
+
+    def __str__(self):  # pragma: nocover
+        return f"Agreement text:\n\n{self.text}"
```
davisagli: This should be truncated if the admin doesn't take care of that automatically. The agreement text will likely be multiple paragraphs.

metadeploy/api/models.py
```diff
@@ -502,6 +502,13 @@ def __str__(self):
         return f"Step {self.name} of {self.plan.title} ({self.step_num})"
 
 
+class ClickthroughText(models.Model):
+    text = models.TextField()
+
+    def __str__(self):  # pragma: nocover
```
davisagli: Is there a good reason NOT to write a test for this?

metadeploy/api/models.py
```diff
@@ -584,6 +594,12 @@ def save(self, *args, **kwargs):
         is_new = self._state.adding
         ret = super().save(*args, **kwargs)
 
+        if is_new:
+            ctt, _ = ClickthroughText.objects.get_or_create(
+                text=self.plan.version.product.click_through_agreement
+            )
+            self.click_through_agreement = ctt
```
jgerigmeyer: Is this actually getting saved when set like this? In the admin, I see the new `ClickthroughText` model created, but the new job still always has `click_through_agreement` as `null` even after accepting the agreement.

metadeploy/api/models.py
wlonk: Actually, I'm not sure what the case for including it at all is. We should display the license text in the admin, that's all.

metadeploy/api/models.py
wlonk: I put it in the wrong place, and the test was getting it off the in-memory object, not out of the database. Thank you for that catch!

metadeploy/api/models.py
```diff
@@ -525,6 +525,10 @@ def __str__(self):
         return f"Step {self.name} of {self.plan.title} ({self.step_num})"
 
 
+class ClickThroughAgreement(models.Model):
+    text = models.TextField()
```
jgerigmeyer: @wlonk Should this be translatable?

davisagli: This is meant to be a record of what the user clicked through. So I think this model should not be translatable, but we should record the text of the translation that the user saw (which I think happens implicitly already).

jgerigmeyer: Yes, correct! Thank you.

PR #220, opened 2019-02-12
[encrypt-oauth-tokens] 🛢 Encrypt tokens at rest

davisagli:
image

This adjusts our oauth code to make sure that tokens are encrypted for storage in the database.

Note the new setting.


app.json
```diff
@@ -78,4 +91,4 @@
       }
     }
   }
-}
+}
```
wlonk: Our editors have different ideas about a file-ending newline, and this is gonna be a change back and forth each time we edit this file 😆

env.example
```diff
@@ -26,3 +26,7 @@ export CONNECTED_APP_CLIENT_SECRET=...
 export CONNECTED_APP_CALLBACK_URL=...
 export CONNECTED_APP_CLIENT_ID=...
 export GITHUB_TOKEN=...
+
+# Change this to a new key,
+# generated using cryptography.fernet.Fernet.generate_key()
+export DB_ENCRYPTION_KEY="Ul-OySkEawSxUc7Ck13Twu2109IzIFh54C1WXO9KAFE="
```
wlonk: I'd suggest defaulting this to `...`, like the secrets above.

metadeploy/multisalesforce/views.py
```diff
@@ -19,15 +21,24 @@
...
 logger = logging.getLogger(__name__)
+FERNET = Fernet(settings.DB_ENCRYPTION_KEY)
+
+
+def fernet_encrypt(s):
+    return FERNET.encrypt(s.encode("utf-8")).decode("utf-8")
+
+
+def fernet_decrypt(s):
+    return FERNET.decrypt(s.encode("utf-8")).decode("utf-8")
```
wlonk: I'm not convinced these should live in `views`; that seems likely to make some connective scar tissue between modules. Maybe abstract them into their own small module, `encryption_utils` or something.

metadeploy/multisalesforce/views.py
```diff
@@ -69,18 +82,26 @@ def complete_login(self, request, app, token, **kwargs):
         ret.account.extra_data[ORGANIZATION_DETAILS] = org_details
         return ret
 
+    def parse_token(self, data):
+        # Encrypt tokens for storage in database
+        data["access_token"] = fernet_encrypt(data["access_token"])
+        data["refresh_token"] = fernet_encrypt(data["refresh_token"])
+        return super().parse_token(data)
```
wlonk: Is this a method that already exists in the Allauth adapters? I don't see it used anywhere else. Maybe add a docstring with a pointer to the relevant Allauth docs.

metadeploy/multisalesforce/views.py
wlonk: This is the highest priority comment from me, the only one that made me hesitate to 👍 this PR.

env.example
davisagli: The Fernet class complains unless you give it a key in the expected format (a URL-safe base64-encoded 32-byte key) and I wanted to keep the existing property of being able to run the app for development using the vars in env.example

app.json
davisagli: My editor disagrees with me about that, so I'll see if we can work out our differences.

metadeploy/multisalesforce/views.py
davisagli: Thanks, I actually intended to ask you explicitly about where these should go.

metadeploy/multisalesforce/views.py
davisagli: It's in the base OAuth2Adapter and called from the oauth2 callback view. I'll add a docstring.

env.example
cdcarter: how about some best of both worlds-ness? I don't love the very required format for the env var, it'll be hard to use the `secret` function in app.json, for example.

cryptography comes with an implementation of HKDF, a key derivation function. with a constant salt in the app, the setting can be of any length and you can derive a fixed length key

config/settings/base.py
```diff
@@ -85,6 +85,7 @@ def env(name, default=NoDefaultValue, type_=str):
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = env("DJANGO_SECRET_KEY")
 HASHID_FIELD_SALT = env("DJANGO_HASHID_SALT")
+DB_ENCRYPTION_KEY = env("DB_ENCRYPTION_KEY")
```
cdcarter: a type coercer here might be helpful too

metadeploy/api/models.py
```diff
@@ -146,7 +148,7 @@ def token(self):
         account = self.social_account
         if account and account.socialtoken_set.exists():
             token = self.social_account.socialtoken_set.first()
-            return (token.token, token.token_secret)
+            return (fernet_decrypt(token.token), fernet_decrypt(token.token_secret))
```
cdcarter: so you're always and forever going to only get a token through the user model, right? :)

config/settings/base.py
wlonk: I think that the type is `str`, which is the default for `env`?

env.example
davisagli: I like that idea, but since this has now been deployed it would require a separate migration to re-encrypt data with the new key.

metadeploy/api/models.py
davisagli: I would have preferred to use https://pypi.org/project/django-encrypted-model-fields/ on the models that actually store the tokens, but since those models are in django-allauth this seemed like the way to get it done.

config/settings/base.py
cdcarter: yea, I'm going to double down on my suggestion of using KDF which lets this just be a str. otherwise i'd probably want to coerce the b64 encoding at least

PR #221, opened 2019-02-13
Make sure we have a valid org id

davisagli:


metadeploy/multisalesforce/views.py
```diff
@@ -20,6 +22,7 @@
 )
 
 logger = logging.getLogger(__name__)
+ORGID_RE = re.compile(r"^00D[a-zA-Z0-9]{15}$")
```
wlonk: This checks for 18-long org IDs, but 15-long ones also exist, right? Are those valid?

wlonk: (`belvedere_utils` contains some logic around that, may be worth combining here?)

davisagli: The UserInfo endpoint gives us an 18-char id, so I figured we might as well not accept anything else.

`belvedere_utils` has functions for converting 15-char to 18-char and for obscuring ids in text, but not for validating an id. If we need this elsewhere at some point we can move it to `belvedere_utils` then.

PR #223, opened 2019-02-14
Fix a few things that were missed re translations and plan templates in the Admin UI and API

davisagli:


metadeploy/api/models.py
```diff
@@ -377,6 +378,9 @@ def preflight_message_markdown(self):
     def post_install_message_markdown(self):
         return self.get_translation("en-us").post_install_message_markdown
 
+    def __str__(self):
+        return self.name
+
```
jgerigmeyer: This is a bit unfortunate when PlanTemplates don't have `name`s by default...

screenshot 2019-02-14 14 15 19

davisagli: Hmm, is that left from a migration before I asked Kit to make plan templates optional? On my local instance there's only one PlanTemplate, which I gave a name.

jgerigmeyer: Oh, I think it's from the `populate_data` script. I can make that provide a name.

davisagli: Thanks -- sorry I missed that.

PR #226, opened 2019-02-20
[use-template-helpers] 📦 Use sfdo-template-helpers in adminapi

wlonk:
kitten


wlonk:
@davisagli if you can do a little functional testing on this, I'd appreciate that, since you have a better sense of plausible uses of the admin API.

PR #229, opened 2019-02-21
Handle no primary plan (and fix spelling)

davisagli:
@jgerigmeyer can you help me with the test for the new branch?

### DIFF
--- a/metadeploy/api/models.py
+++ b/metadeploy/api/models.py
@@ -13,7 +13,7 @@ from django.conf import settings
 from django.contrib.auth.models import AbstractUser
 from django.contrib.auth.models import UserManager as BaseUserManager
 from django.contrib.postgres.fields import JSONField
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.validators import RegexValidator
 from django.db import models
 from django.db.models import Count, F, Func, Q
@@ -351,8 +351,10 @@ class Version(HashIdMixin, TranslatableModel):
 
     @property
     def primary_plan(self):
-        # This will raise an error if the number of primary plans != 1:
-        return self.plan_set.filter(tier=Plan.Tier.primary).get()
+        try:
+            return self.plan_set.filter(tier=Plan.Tier.primary).get()
+        except ObjectDoesNotExist:
+            return None
 
     @property
     def secondary_plan(self):
--- a/src/js/components/products/detail.js
+++ b/src/js/components/products/detail.js
@@ -126,7 +126,8 @@ class VersionDetail extends React.Component<VersionDetailProps> {
                   {t('Select a Plan to Install')}
                 </h3>
                 <p>{version.description}</p>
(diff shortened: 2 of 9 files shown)


src/js/components/products/detail.js
```diff
@@ -126,7 +126,8 @@ class VersionDetail extends React.Component<VersionDetailProps> {
                   {t('Select a Plan to Install')}
                 </h3>
                 <p>{version.description}</p>
-                {version.primary_plan.is_listed &&
+                {version.primary_plan &&
```
jgerigmeyer: @davisagli I'm curious about this change. My understanding was that versions were _required_ to have one (and only one) `primary_plan`, and this was also enforced in the serializer. See:

https://github.com/SFDO-Tooling/metadeploy/blob/master/metadeploy/api/models.py#L352-L355
https://github.com/SFDO-Tooling/metadeploy/blob/master/metadeploy/api/tests/models.py#L132-L135

Has this expectation changed?

davisagli: Hmm. Now I'm puzzled why things seemed to work when I made this change locally.

Here's my scenario:
- product with an existing primary plan
- I want to turn the existing primary plan into an additional plan
- There will be a new primary plan but I'm not ready to publish it yet

jgerigmeyer: I buy that... I'm not sure there's a _good_ reason to enforce a `primary_plan`, but we should probably adjust the back-end accordingly.

davisagli: Yeah, I agree. I'm in a meeting now but let's hold this until we have a chance to work on that part too.

jgerigmeyer: I already have it adjusted locally, so I'll push shortly and then we can discuss/adjust.

metadeploy/api/models.py
```diff
@@ -351,8 +351,10 @@ def __str__(self):
 
     @property
     def primary_plan(self):
-        # This will raise an error if the number of primary plans != 1:
-        return self.plan_set.filter(tier=Plan.Tier.primary).get()
+        try:
+            return self.plan_set.filter(tier=Plan.Tier.primary).get()
+        except ObjectDoesNotExist:
+            return None
```
davisagli: You can use `first()` to avoid needing to catch the exception: https://docs.djangoproject.com/en/2.1/ref/models/querysets/#first

jgerigmeyer: Right, this was intentional. I figured we _do_ want to throw a potential `MultipleObjectsReturned` exception?

https://docs.djangoproject.com/en/2.1/ref/models/querysets/#get

davisagli:
@jgerigmeyer I can't approve my own PR but it looks good to me.

metadeploy/api/models.py
wlonk: I see you all had the conversation I was going to have here. Yes, we do want to be alerted to the undesirable and out-of-spec scenario where we have `MultipleObjectsReturned`.

PR #237, opened 2019-02-26
[install-progress] 🐈 Continuous install progress.

jgerigmeyer:
https://trello.com/c/e0gK46gk

caracal


src/js/components/jobs/progressBar.js
```diff
@@ -7,85 +7,174 @@ import { t } from 'i18next';
...
+    }
+  }
+
+  static getCompletedSteps(job: JobType): number {
+    // Get number of completed steps
+    return job.steps.filter(
+      step =>
+        job.results[step] &&
+        job.results[step].find(
+          res => res.status === CONSTANTS.RESULT_STATUS.OK,
```
davisagli: I don't think we're currently using this but note that steps have an `allow_failure` boolean and if true, it'll proceed to the next step after an error.

jgerigmeyer: Huh, interesting. Yeah, that's not accounted for in MetaDeploy as of yet. I think that'll be its own story.

jgerigmeyer:
@mirisuzanne I'd appreciate your functional review on this PR. After doing the `python manage.py truncate_data && python manage.py populate_data` dance, you should be able to run an install and see the new continuous-progress on the installer.

PR #238, opened 2019-02-26
[ux-cleanup] 🐈 Various minor ux fixes (from user testing).

jgerigmeyer:
https://trello.com/c/uxZtPKgc
https://trello.com/c/z0JQnmrI
https://trello.com/c/wNu8kotB
https://trello.com/c/JpXBS22X
https://trello.com/c/LWHhS3gf
https://trello.com/c/02T6vVvf
https://trello.com/c/29d2fjTk
https://trello.com/c/W6l1qhkP

zebras


jgerigmeyer:
@mirisuzanne @SondraE I'd appreciate functional review on this. It includes minor changes from 7 different Trello stories. Sorry, I got kinda carried away... 😬

metadeploy/api/management/commands/populate_data.py
```diff
@@ -239,7 +239,9 @@ def handle(self, *args, **options):
         self.create_enqueuer_job()
         self.create_token_expiry_job()
         self.create_preflight_expiry_job()
-        sf_category = ProductCategory.objects.create(title="salesforce", order_key=0)
+        sf_category = ProductCategory.objects.create(
+            title="salesforce.org", order_key=0
+        )
```
jgerigmeyer: @davisagli One of the UX recommendations was to use "salesforce.org" instead of "salesforce" for the product category. Obviously up to you all whether that's a good idea, and worth a data migration.

src/js/components/plans/toasts.js
```diff
@@ -38,7 +38,6 @@ class Toasts extends React.Component<Props, State> {
           heading: [t(label)],
         }}
         variant={variant}
-        duration={20 * 1000}
```
jgerigmeyer: I'll note one thing I don't love about this (maybe we can fix with overriding default SLDS styles?)... Toasts prevent clicking on the entire header, which means users can't login/logout or click the logo until dismissing the toast.

Not directly related to this change, I guess -- probably should be its own bug story.

metadeploy/api/management/commands/populate_data.py
davisagli: We don't yet have multiple product categories configured so this isn't visible, but I went ahead and updated it on production for when it is.

src/js/components/products/detail.js
```diff
@@ -142,7 +140,7 @@ class VersionDetail extends React.Component<VersionDetailProps> {
                         slds-button_brand
                         slds-size_full"
                     >
-                      {primary_plan.title}
+                      {t('View')} {primary_plan.title}
```
davisagli: Just to confirm, we are meant to manually rename existing plans so that they do not have a verb in their title?

I'm not convinced this will resolve the confusion (the heading says "Select a plan to install" but "View" is a different verb than "Select" or "install") but we can try it.

src/js/components/plans/toasts.js
jgerigmeyer: It scrolls with the screen, so it gets in the way of whatever is at the top of your viewport (left/right of the toast).

screenshot 2019-02-27 14 32 28

I added a card: https://trello.com/c/ZGlV9fQT

davisagli: Got it. I agree that's annoying but I don't think it's very high priority to fix.

locales/en/translation.json
```diff
@@ -73,7 +73,10 @@
   "Run pre-install validation": "Run pre-install validation",
   "Salesforce Org Information": "Salesforce Org Information",
   "Sandbox or Scratch Org": "Sandbox or Scratch Org",
-  "Select a Plan to Install": "Select a Plan to Install",
+  "Select a Plan": "Select a Plan",
+  "Select a Product to Install": "Select a Product to Install",
+  "Select a different plan": "Select a different plan",
+  "Select a different product": "Select a different product",
```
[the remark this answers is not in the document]
jgerigmeyer: Yeah, it's a distinction between headings and body-text/links.

src/js/components/products/detail.js
jgerigmeyer: The heading now says "Select a Plan".

I'm not sure about plan titles with verbs. Can you provide some examples of existing plan titles?

davisagli: Current plans on production:
screen shot 2019-02-27 at 2 43 02 pm

jgerigmeyer: @davisagli Ah, that's helpful -- I didn't realize how they were currently worded.

The "Install" ones seem redundant to me, and "Install" could be removed. But "Load Sample Data" makes sense as-is... I'd say if we want to go with this "View" change, we maybe add quotes around the plan title? E.g. `View "Load Sample Data"`

cc @mirisuzanne @SondraE @stacyk

davisagli: What about just showing the plan title as it is configured, but adding a line below the heading that says something like "You will have a chance to confirm the steps of the plan before it starts."

jgerigmeyer: I'm okay with either of those options, but I'd defer to @mirisuzanne @SondraE @stacyk.

CONTRIBUTING.rst
```diff
@@ -42,6 +42,9 @@ two different arbitrary string values. Also set ``DB_ENCRYPTION_KEY``::
     from cryptography.fernet import Fernet
     Fernet.generate_key()
 
+This will output a string, e.g. ``b'mystring='``. Copy just the contents of
```
wlonk: Pedantically, a bytestring. Not sure that I think the distinction matters here though.

PR #241, opened 2019-03-01
[plan-permalink] Move plan slugs to plan templates

wlonk:
cat


metadeploy/api/models.py
```diff
@@ -404,20 +392,45 @@ class PlanSlug(models.Model):
     def validate_unique(self, *args, **kwargs):
         super().validate_unique(*args, **kwargs)
         qs = PlanSlug.objects.filter(
-            parent__version=self.parent.version, slug=self.slug
+            parent__plan__version__product__in=self.get_associated_products(),
```
davisagli: We would like PlanTemplate to have a direct relation to Product. That is partly motivated because the plan publishing task in CumulusCI needs to be able to find an existing PlanTemplate knowing the product and PlanTemplate name, even if there aren't any plans yet. But it would also simplify the code here, because we can just validate that the slug is unique among all PlanTemplates for the product, and separately on Plan validate that (version, plantemplate) is a unique combo.

src/js/components/products/detail.js
```diff
@@ -76,6 +76,23 @@ class VersionDetail extends React.Component<VersionDetailProps> {
...
+      ) {
+        const slugs = [
+          product.most_recent_version.primary_plan.slug,
+          product.most_recent_version.secondary_plan.slug,
+          ...product.most_recent_version.additional_plans.map(e => e.slug),
+        ];
+        if (slugs.includes(versionLabel)) {
+          return; // Redirect routes.plan_detail(product.slug, product.most_recent_version.slug, versionLabel)
+        }
+      }
```
jgerigmeyer: @wlonk The logic here is on the right track, but there are some complexities around wanting to do this check _before_ making an extra fetch to the API to see if the slug is an older version label. I'm happy to talk through those together tomorrow if you're willing.

wlonk: Sounds good. I also wasn't sure how to short circuit to rendering a redirect here.

wlonk:
@davisagli it's my go-to when cornered in a situation like that.

In particular, `.update(foo=F('bar__baz'))` doesn't work (can't update an F with a join apparently), and explicitly loading the objects into memory and saving them doesn't work because of parler not working right in migration files with the faked-out models.

src/js/store/products/selectors.js
```diff
@@ -77,11 +82,59 @@ const selectVersion: (
...
+    maybeVersionLabel: ?string,
+  ): VersionPlanType => {
+    // There's a chance that the versionLabel is really a planSlug.
+    // In that case, check the most recent version in the product and see.
+    if (
+      product &&
+      product.most_recent_version &&
+      product.most_recent_version.primary_plan &&
+      product.most_recent_version.secondary_plan &&
+      product.most_recent_version.additional_plans
```
jgerigmeyer: This seems overly specific... don't we also want to follow this branch if there's a `primary_plan` but not a `secondary_plan`?

src/js/store/products/selectors.js
```diff
@@ -77,11 +82,59 @@ const selectVersion: (
...
+  InitialProps,
+) => VersionPlanType = createSelector(
+  [selectProduct, selectVersionLabel],
+  (
+    product: ProductType | null,
+    maybeVersionLabel: ?string,
+  ): VersionPlanType => {
+    // There's a chance that the versionLabel is really a planSlug.
+    // In that case, check the most recent version in the product and see.
+    if (
```
jgerigmeyer: Less nesting if we bail early (similar to the other selectors):

```js
    if (!product || !maybeVersionLabel) {
      return { label: null, slug: null };
    }
```

src/js/components/products/detail.js
```diff
@@ -86,22 +96,23 @@ class VersionDetail extends React.Component<VersionDetailProps> {
...
+      versionLabelAndProductSlug !== prevProps.versionLabelAndProductSlug;
     if (versionChanged) {
       this.fetchVersionIfMissing();
     }
   }
 
   render(): React.Node {
-    const { user, product, version, versionLabel } = this.props;
+    const { user, product, version, versionLabelAndProductSlug } = this.props;
+    const { label } = versionLabelAndProductSlug;
```
jgerigmeyer: I'm not seeing the actual `<Redirect>` anywhere. I would expect it around here... although I guess it maybe fits cleaner into the `getLoadingOrNotFound` fn in `components/utils.js`.

I'm imagining adding a `planSlug` (optional) argument, then something like:

```js
  if (product === null) {
    return <ProductNotFound />;
  }
  if (planSlug && versionLabel && !plan) {
    return <Redirect to={routes.plan_detail(product.slug, versionLabel, planSlug)} />;
  }
  ...
```

src/js/components/products/detail.js
```diff
@@ -35,7 +35,10 @@ type VersionDetailProps = {
   user: UserType,
   product: ProductType | null,
   version: VersionType | null,
-  versionLabel: ?string,
+  versionLabelAndProductSlug: {
```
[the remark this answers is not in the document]
wlonk: THAT'S WHY EVERYTHING WAS FAILING.

Apparently it is hard for me to spot the difference between "plan" and "product" reliably.

src/js/components/products/detail.js
wlonk: That's because I hadn't gotten to adding the redirect—I wanted to make sure that I could swap in the label-or-slug logic first. But I'll throw this in.

src/js/store/products/selectors.js
wlonk: Ugh, yes, we do. I put it this way because I then assemble the list of slugs assuming all of these have values.

src/js/components/products/detail.js
```diff
@@ -123,6 +123,9 @@ class VersionDetail extends React.Component<VersionDetailProps> {
     if (!product || !version) {
       return <ProductNotFound />;
     }
+    if (slug && label) {
+      return <Redirect to={routes.plan_detail(product.slug, label, slug)} />;
+    }
```
jgerigmeyer: @wlonk Sorry, I wasn't clear in my comment. I think this logic (or what I pasted in my comment) needs to live in the `getLoadingOrNotFound` fn in `components/utils.js`. Otherwise that will trigger first, and we'll get a `<VersionNotFound />` instead of the redirect.

src/js/store/products/selectors.js
```diff
@@ -12,6 +12,11 @@ import type {
 
 export type ProductsMapType = Map<string, Array<ProductType>>;
 
+type VersionPlanType = {
+  +label: string,
+  +slug: string,
+};
```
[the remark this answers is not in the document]
jgerigmeyer: For consistency elsewhere, let's use `string | null`? I don't think we expect `undefined`.

src/js/components/products/detail.js
wlonk: Oh! Gotcha.

jgerigmeyer:
@davisagli I think this is ready to go, once you've had a chance to review it and/or test it out functionally.

davisagli:
I tested this locally. The migration ran as expected, but the button to go from the product page to the plan page is linking to http://localhost:8080/products/npsp/3.149/null even though the Plan is associated with a PlanTemplate that has a slug.

jgerigmeyer:
> I tested this locally. The migration ran as expected, but the button to go from the product page to the plan page is linking to http://localhost:8080/products/npsp/3.149/null even though the Plan is associated with a PlanTemplate that has a slug.

@wlonk can you take a look at this?

wlonk:
@jgerigmeyer yes, it's enqueued, I'm just working on the logging issues first :)

wlonk:
I'm guessing that the slug is not being passed through from its new location to the serializer correctly.

wlonk:
@davisagli great to hear; can you clarify what the misconfiguration was, so that we can spot it and better avoid it if someone else lands there?

davisagli:
@wlonk I thought I had created the slug but I hadn't, simple as that 😒

wlonk:
Hah, well, that'd do it!

PR #245, opened 2019-03-07
Fix products admin API

davisagli:
django-filters doesn't know what to do with an ImageField


davisagli:
@wlonk yes in theory, but I'd call it very low priority.

davisagli:
Passing now, needs re-approval

PR #249, opened 2019-03-07
[improve-background-task-logging] Show error tracebacks from background jobs

wlonk:
cats

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -407,7 +407,11 @@ LOGGING = {
             "propagate": False,
         },
         "django.server": {"handlers": ["console"], "level": "INFO", "propagate": False},
-        "django.request": {"handlers": ["console_error"], "level": "INFO"},
+        "django.request": {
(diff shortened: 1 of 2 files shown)


davisagli:
As I said in slack, the current capturing of tracebacks for failed django-rq jobs is fine, so I don't think we need/want this change, at least for now. We do need tracebacks for exceptions in Django views to get logged though.

wlonk:
@davisagli I seem to be getting full stacktraces in my console now; do you want to take a look and see if it does what you need?

davisagli:
@wlonk I tried it -- now I see 2 copies of each traceback in the console

wlonk:
@davisagli Hm, I'm getting that too. And it is completely unclear to me _why_. I will dig.

davisagli:
@wlonk looks like I only get it once if I put `"propagate": False` on the `django.request` logger

wlonk:
@davisagli can confirm, pushed that.

PR #250, opened 2019-03-08
[choose-yarn-node] 🙀 Pull node and yarn versions from package.json

wlonk:
Our bash scripts have no testing infrastructure, so I hand-tested this.

cat


wlonk:
Once we like this, we can backport it into https://github.com/SFDO-Tooling/sfdo-template/pull/10

bin/unpack-node
```diff
@@ -13,11 +13,40 @@ else
...
 fi
 
-nodefile="../requirements/node-v10.15.0-${platform}-x64.tar.xz"
-yarnfile="../requirements/yarn-v1.13.0.tar.gz"
+
+function get_from_package {
+    python << EOF
+import json
+with open('../package.json') as f:
+    out = json.loads(''.join(f.readlines())).get('engines', {}).get('$1', '')
```
davisagli: I think you can use `json.load` with a file object instead of json.loads with a string.

PR #253, opened 2019-03-11
[site-object] 🛢🐈 Add basic site profile object.

wlonk:
See https://trello.com/c/JG5FMRPZ/132-i-want-a-configurable-non-product-related-deployment-of-metadeploy

cat


wlonk:
@jlantz @jgerigmeyer This is the start of what we need for https://trello.com/c/JG5FMRPZ/132-i-want-a-configurable-non-product-related-deployment-of-metadeploy

But I don't know what the full version of that story looks like. Can some of us get together and talk about it in the not-too-distant future and figure out what the acceptance criteria for this might be? Thanks!

davisagli:
@wlonk I think this is heading in the right direction.

name should be translated.

A separate installation would also potentially need the ability to customize the copyright notice.

wlonk:
@davisagli should name be translated? It didn't seem to me that it would vary per-language.

Customized copyright notice is good.

@jgerigmeyer do you want to query this object on the frontend, to populate those text areas? I can make an endpoint to that.

davisagli:
@wlonk I'm thinking of a site name that is like `[brand] Installer`

wlonk:
@davisagli ah ha, got it.

jgerigmeyer:
@wlonk This looks like a good start. Rather than adding an API query, can we just pass this data through the `GLOBALS` var that gets injected into the index.html template?

wlonk:
@jgerigmeyer much better idea! I will do it.

jgerigmeyer:
@wlonk Looks good to me. Go ahead and make the front-end adjustments as well?

metadeploy/api/migrations/0063_populate_site_profile.py
```diff
@@ -0,0 +1,48 @@
...
+    activate("en")
+    SiteProfileTranslation.get_translated_fields = get_translated_fields
+    SiteProfile._parler_meta = ParlerOptions(
+        None,
+        shared_model=SiteProfile,
+        translations_model=SiteProfileTranslation,
+        related_name=compat.get_remote_field(SiteProfileTranslation).related_name,
+    )
+    for site in Site.objects.all():
+        SiteProfile.objects.create(site=site, name=site.name)
```
jgerigmeyer: This migration doesn't seem to be working as I'd expect. I always get a SiteProfile with blank `name`, even though my Site object does have a name.

src/js/components/header/index.js
```diff
@@ -32,12 +32,22 @@ const Header = ({ user, doLogout, socket }: Props) => (
...
         >
-          <span data-logo-bit="start">meta</span>
-          <span data-logo-bit="end">deploy</span>
+          {window.GLOBALS.SITE && window.GLOBALS.SITE.logo ? (
+            <img
+              className="slds-size_small"
+              src={window.GLOBALS.SITE.logo}
+              alt={window.SITE_NAME}
+              title={window.SITE_NAME}
+            />
```
jgerigmeyer: @mirisuzanne Do we want to restrict the height here, or leave that up to the person adding the logo?

Screenshot 2019-03-14 11 13 06

metadeploy/api/migrations/0063_populate_site_profile.py
wlonk: Probably because of faked out translation issues.

wlonk: (Addendum this is a reason for getting pytedjmi working and robust! Test these migrations.)

src/js/components/header/index.js
mirisuzanne: We should restrict it.

metadeploy/api/migrations/0063_populate_site_profile.py
wlonk: Confirmed, I was using the wrong language code (`"en"` instead of `"en-us"`). Roll this back and forward and you should see it fixed.

metadeploy/api/models.py
```diff
@@ -762,3 +763,26 @@ def save(self, *args, **kwargs):
...
+class SiteProfile(TranslatableModel):
+    site = models.OneToOneField(Site, on_delete=models.CASCADE)
+
+    translations = TranslatedFields(
+        name=models.CharField(max_length=64),
+        welcome_text=MarkdownField(property_suffix="_markdown", blank=True),
+        copyright_notice=MarkdownField(property_suffix="_markdown", blank=True),
+    )
+
+    logo = models.ImageField(blank=True)
```
jgerigmeyer: @wlonk Based on our conversation, I think what's missing here is a `company_logo` field or something like that? I'd be tempted to add yet another field `favicon`, even if it falls back to `company_logo`.

src/js/components/footer.js
```diff
@@ -26,7 +26,16 @@ const Footer = (props: { logoSrc: string }) => (
...
-      <p>{t('Copyright 2000–2019 Salesforce.org. All rights reserved.')}</p>
+      {window.GLOBALS.SITE && window.GLOBALS.SITE.copyright_notice ? (
+        // These messages are pre-cleaned by the API
+        <div
+          dangerouslySetInnerHTML={{
+            __html: window.GLOBALS.SITE.copyright_notice,
+          }}
+        />
+      ) : (
+        <p>{`${t('Copyright')} ${window.GLOBALS.YEAR} ${window.SITE_NAME}.`}</p>
```
jgerigmeyer: @wlonk I'm wondering if we want to add a `company_name` field? Seems like it's likely distinct from the product name.

We could use it here in the default copyright, and also in the `<meta name="author">` tag I removed.

jgerigmeyer:
@davisagli Just a heads-up that this removes "Salesforce.org" and the SFDO logo by default, so you'll want to edit the new site-object on any deployments. I left the `salesforce-logo.png` and `salesforce-favicon.png` files in the repo (`static/images/`) to make that easier, even tho they're not being used directly.

jgerigmeyer:
@davisagli As far as I'm aware this is ready to merge once you give your 👍.

davisagli:
@jgerigmeyer I saw this; was busy today but will try to get it merged and deployed tomorrow.

jgerigmeyer:
> @jgerigmeyer I saw this; was busy today but will try to get it merged and deployed tomorrow.

No rush; just wanted to be sure you were aware.

PR #254, opened 2019-03-14
[checkbox-clarity] Remove disabled checkboxes.

jgerigmeyer:
https://trello.com/c/5ggjFz9B

cuy

### DIFF
--- a/src/js/components/plans/stepsTable/index.js
+++ b/src/js/components/plans/stepsTable/index.js
@@ -58,6 +58,8 @@ const StepsTable = ({
       }
     }
   }
+  const hasValidToken = user && user.valid_token_for !== null;
+  const hasReadyPreflight = preflight && preflight.is_ready;
   return (
     <div
       className="slds-p-around_medium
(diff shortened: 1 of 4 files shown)


jgerigmeyer:
@SondraE @mirisuzanne This is pretty straightforward for the plan-detail page (check icons when required, dash icons when skipped), but I wasn't sure what to do on the install-running page. What icons make sense for "waiting to install" vs "will not install / skipped"?

Screenshot 2019-03-14 11 56 37

jgerigmeyer:
> I wonder if we should run the checkmark color by their accessibility team? We have the label for those who use screen readers, but for those who don't but just have some lower contrast screens, I am curious to find out if checkmark is too light?
> 
> Code looks good from what I can tell :)

We are using the lighter icon color, and could easily switch that to a darker grey.

jgerigmeyer:
@stacyk So on the install page, you're okay with the checkmark icon representing "waiting to install" and the dash icon representing both "skipped" and "will not install" / "cancelled"?

stacyk:
> @stacyk So on the install page, you're okay with the checkmark icon representing "waiting to install" and the dash icon representing both "skipped" and "will not install" / "cancelled"?

I am good with the icon and what they represent, and I am good with the color after confirming this is noted in the SLDS docs: https://www.lightningdesignsystem.com/components/icons/#Light

PR #257, opened 2019-03-18
[narrow-toast] Add max-width to toasts.

stacyk:
Trello: https://trello.com/c/ZGlV9fQT/129-bug-cannot-click-on-header-login-logout-logo-when-toast-is-visible

### DIFF
--- a/src/js/components/plans/toasts.js
+++ b/src/js/components/plans/toasts.js
@@ -100,7 +100,11 @@ class Toasts extends React.Component<Props, State> {
 
   render(): React.Node {
     const { isOpen } = this.state;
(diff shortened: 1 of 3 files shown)


stacyk:
Screenshot with new CSS edits shown:
half-toast-container

stacyk:
> This is not a huge deal, but it gets a bit weird / off-center and cut-off on smaller screen sizes.
> 
> Screenshot 2019-03-18 14 07 17

The toast has a min-width of 30rem so I suppose I can make that the max-width as well and we would get a bit more space than we currently do. I can give it a whirl quick and see how that works.

stacyk:
@jgerigmeyer I made some edits to the width of the toast at different screen sizes. Check it out when you have a second.

PR #258, opened 2019-03-18
[install-logs] Show logs as they come in

wlonk:
Required:

- `./manage.py truncate_data`
- `./manage.py populate_data`

See https://trello.com/c/pnXbwwSW/101-as-an-admin-i-want-to-view-the-logs-while-running-a-preflight-installation

maine coon


wlonk:
@davisagli can you help us get real-er tasks running so we can see the sort of data we'd really get as they run?

davisagli:
@wlonk if I publish the actual installation plan for one of our products to metadeploy-devstaging, is that what you're looking for?

wlonk:
@davisagli I think so? What I would love in my perfect imagined state of the world is a task we can run that acts 100% like a real plausible task except it doesn't actually modify a Salesforce org, so we can run it repeatedly for testing. This could get us there?

davisagli:
@wlonk hmm. The tasks we use for real were created with the express purpose of modifying Salesforce orgs, so that's a pretty big "except". Is there a reason it doesn't work to run a real installation in a scratch org, then delete and re-create the scratch org for the next run?

Alternately, for testing this if it's more important to have a task that logs things over time than the specific output from our real tasks, we could pretty easily make a task that logs the same message once per second for a configured number of seconds.

wlonk:
@davisagli There's a reason, but it's not a good one: that's not integrated into my workflow, and I don't know how to do that easily.

Your "alternatively" sounds spot-on. That could be fine.

wlonk:
OK, this now has actual log data, but it's not broken up by task. @jgerigmeyer I'd love to talk with you about how the frontend will consume this and what I can do to make that easier for you.

metadeploy/api/tests/serializers.py
```diff
@@ -261,9 +261,10 @@ def test_invalid_steps_made_valid_by_preflight(
         assert serializer.is_valid(), serializer.errors
 
     def test_no_context(self, job_factory):
-        job = job_factory()
+        job = job_factory(status=Job.Status.complete, results={"log": "===="})
```
jgerigmeyer: Is this "log" or "logs"?

https://github.com/SFDO-Tooling/MetaDeploy/pull/258/files#diff-3abadce051d8e41a9a9531194a508996R13

jgerigmeyer:
> OK, this now has actual log data, but it's not broken up by task. @jgerigmeyer I'd love to talk with you about how the frontend will consume this and what I can do to make that easier for you.

@wlonk Let's see... The short version is that the JS will just need the logs (are they plain strings, or markdown?) broken up by step-id. I assume you'll be pushing them over websockets. Right now we're using a `JOB_STEP_COMPLETED` event... I'm guessing we'll want to add a new (debounced/throttled?) event for logging during installation (e.g. `JOB_STEP_LOGS_UPDATED` or similar). Happy to discuss further when you're available.

wlonk:
@davisagli I've been looking into this, and I think the only way to get the logs divided by task is with a change to `cumulusci.core.flowrunner`, to pass in the current step to the logger calls. Is that a thing you can do?

davisagli:
@wlonk why can't you do what you need to do in the callbacks in metadeploy/api/flows.py? It already sets up its own logging handler for the flow, and you can use the `pre_task`/`post_task` callbacks to keep track of the step context.

wlonk:
@davisagli ah ha, we can! I got that working. Thanks.

wlonk:
@mirisuzanne or @stacyk if you want to take this and make the styles work, I think it's ready for you! It may be super annoying to do as thing flow through, so just wait until a job is done and then open an accordion and you can see how it looks as it flows.

metadeploy/api/flows.py
```diff
@@ -42,26 +43,41 @@ def pre_flow(self, coordinator):
...
+                self.context.results[step_id] = {}
             if result.exception:
-                self.context.results[step_id] = [
+                self.context.results[step_id].update(
                     {"status": ERROR, "message": bleach.clean(str(result.exception))}
-                ]
+                )
             else:
-                self.context.results[step_id] = [{"status": OK}]
+                self.context.results[step_id].update({"status": OK})
```
jgerigmeyer: @wlonk Does this only apply to jobs, or also to preflights? I assume preflights are unchanged in the branch, and still have a syntax `results: { 'this-step': { status: 'ok' } }`?

metadeploy/api/result_spool_logger.py
```diff
@@ -0,0 +1,21 @@
...
+    def emit(self, record):
+        try:
+            msg = f"\n{record.getMessage()}"
+            self.result.results[self.current_key]["logs"] += msg
+        except KeyError as e:
+            msg = record.getMessage()
+            if e.args[0] == "logs":
+                self.result.results[self.current_key]["logs"] = msg
+            else:
+                self.result.results[self.current_key] = {"logs": msg}
```
jgerigmeyer: What's all this? Maybe add an explanatory comment?

src/js/components/plans/jobResults.js
```diff
@@ -45,38 +45,36 @@ export const WarningIcon = (): React.Node => (
...
-              {/* These messages are pre-cleaned by the API */}
-              <span dangerouslySetInnerHTML={{ __html: err.message }} />
-            </li>
-          );
-      }
-      return null;
-    })}
-  </ul>
-);
+  errorList: StepResultType,
```
jgerigmeyer: This means that a preflight can only have one plan-level error. I think that's okay.

src/js/components/plans/stepsTable/index.js
```diff
@@ -51,12 +51,15 @@ const StepsTable = ({
...
     for (const step of job.steps) {
-      if (!job.results[step]) {
-        activeJobStep = step;
+      if (job.results[step]) {
+        lastSeenStep = step;
+      } else {
         break;
       }
     }
+    activeJobStep = lastSeenStep;
```
jgerigmeyer: @wlonk I don't understand why this change is necessary.

wlonk: The currently running step does, with a `log` attribute as they are running.

wlonk:
@jgerigmeyer that's where the island of misfit logs lives; those are ones that exist outside a step. I wasn't sure yet whether we wanted to show them anywhere.

jgerigmeyer:
> @jgerigmeyer that's where the island of misfit logs lives; those are ones that exist outside a step. I wasn't sure yet whether we wanted to show them anywhere.

My understanding from our mtg with Jason was that we don't care about logs not associated with a specific step.

jgerigmeyer:
@wlonk Would you be willing to do a careful audit of the websocket events? During installation, the front-end is getting hammered with `TASK_COMPLETED` events. Sometimes it's because the `null` logs have changed (which I don't think we want in any case), and other times there are literally no changes in the payload data and we still get ~8 events in a row. I still think debouncing or throttling may be in order.

wlonk:
@jgerigmeyer looks like taking out the misfit logs helps, but mostly it's that each log line triggers a new websocket event, and then any change in status triggers one, too, and it has to be that way. I think with real tasks we might get a different pattern, but the sleeping-fake-tasks are just kinda hammering the frontend. It performant for me, at least, so I'm not too concerned.

jgerigmeyer:
> @jgerigmeyer looks like taking out the misfit logs helps, but mostly it's that each log line triggers a new websocket event, and then any change in status triggers one, too, and it has to be that way. I think with real tasks we might get a different pattern, but the sleeping-fake-tasks are just kinda hammering the frontend. It performant for me, at least, so I'm not too concerned.

Hm, that definitely helps, but does that explain the ~3-4 in a row with no changes in status or logs? I'm getting those consistently, especially around the status change.

BTW, the redux dev tools extension is really helpful for seeing this: http://extension.remotedev.io/

wlonk:
Thanks, I'll try that; I was having trouble visualizing what was going on.

wlonk:
@jgerigmeyer I have that extension, but I wasn't thinking to use it in this case; I was looking at incoming websockets. Anyway, the deal is that this method should be pushing once:

```
    def push_if_results_changed(self):
        results_has_changed = self.tracker.has_changed("results") and self.results != {}
        self._push_if_condition(results_has_changed, notify_post_task)
```

but seems to be pushing repeatedly; my guess is that the tracker isn't up to what we're asking of it.

jgerigmeyer:
> @jgerigmeyer I have that extension, but I wasn't thinking to use it in this case; I was looking at incoming websockets. Anyway, the deal is that this method should be pushing once:
> 
> ```
>     def push_if_results_changed(self):
>         results_has_changed = self.tracker.has_changed("results") and self.results != {}
>         self._push_if_condition(results_has_changed, notify_post_task)
> ```
> 
> but seems to be pushing repeatedly; my guess is that the tracker isn't up to what we're asking of it.

Hm, it seems less-than-ideal to not know why, especially since it could cause issues down the road. But in this immediate I guess it's okay? 🤷‍♂️

wlonk:
@jgerigmeyer Oh, I'm not done—I'm still investigating.

wlonk:
WORKING HYPOTHESIS! It's a race in the server.

So it looks like what's happening is this: the save events are triggering in-order, putting push events on the redis layer to be consumed by the channels consumer. That's all in-order. The channels consumer picks up events in order or not (can't tell, and doesn't quite matter) and re-hydrates the models to send out over the websocket; since it was put on the queue, the model has gained some lines in the logs, and so you get three or so that are the same because they were enqueued as different things, but by the time they're sent, the consumer picks them up as the same.

Now, how to deal with this? That's my next question.

jgerigmeyer:
> WORKING HYPOTHESIS! It's a race in the server.
> 
> So it looks like what's happening is this: the save events are triggering in-order, putting push events on the redis layer to be consumed by the channels consumer. That's all in-order. The channels consumer picks up events in order or not (can't tell, and doesn't quite matter) and re-hydrates the models to send out over the websocket; since it was put on the queue, the model has gained some lines in the logs, and so you get three or so that are the same because they were enqueued as different things, but by the time they're sent, the consumer picks them up as the same.
> 
> Now, how to deal with this? That's my next question.

Great! It's good to at least know what's going on. I still think throttling/debouncing events to a single subscriber is a sensible long-term strategy, but I don't know how difficult that is.

wlonk:
Neither do I, yet. I think what we might do for now is add a backlog ticket with this information and prioritize debouncing as its own story, and let this PR move forward with some excessive websocket activity for now.

duggiemitchell:
> I think this state might need to belong to the parent component though? So it can be shared by this component (which toggles it) and then also passed to the component (which uses it to expand/collapse the logs).

duggiemitchell:
> The first appearance of React Hooks in this project! I love it. 🎉

yolo ¯\_(ツ)_/¯

src/js/components/plans/stepsTable/index.js
```diff
@@ -33,74 +34,119 @@ export type DataCellProps = {
...
-              label={<InstallDataColumnLabel />}
-              property="is_recommended"
+              key="name"
+              label={
+                job ? (
+                  <ToggleLogsDataColumnLabel
+                    showLogs={this.state.showLogs}
+                    toggleLogs={() =>
+                      this.setState({ showLogs: !this.state.showLogs })
+                    }
```
jgerigmeyer: I don't think this is quite right. `showLogs` should become `true` when the user clicks the "Show Logs" button, and `false` when they click "Hide Logs" -- but the state of that button should actually be determined (I think?) by whether there are any steps expanded. In other words, we need to also pass in `expandedPanels` here, and the toggle state can be determined based on whether or not there are any panels expanded.

duggiemitchell:
> > Note: We import React as a namespace here with import * as React from 'react' instead of as a default with import React from 'react'. When importing React as an ES module you may use either style, but importing as a namespace gives you access to React’s utility types.
> 
> https://flow.org/en/docs/react/components/

This makes sense, I was wondering why before I couldn't assign a SFC to something like React.Node.  I think I changed this when attempting to use `useState`.  Would it be best to still import as a namespace like so: 
```
import * as React from 'react';
import { useState } from 'react';
```

jgerigmeyer:
> > > Note: We import React as a namespace here with import * as React from 'react' instead of as a default with import React from 'react'. When importing React as an ES module you may use either style, but importing as a namespace gives you access to React’s utility types.
> > 
> > 
> > https://flow.org/en/docs/react/components/
> 
> This makes sense, I was wondering why before I couldn't assign a SFC to something like React.Node. I think I changed this when attempting to use `useState`. Would it be best to still import as a namespace like so:
> 
> ```
> import * as React from 'react';
> import { useState } from 'react';
> ```

Hm, I haven't played with it, though I'd think using it as `React.useState` wouldn't be too bad.

src/js/components/plans/stepsTable/index.js
duggiemitchell: Yup think that's right and should there also be some state that disables the toggle if there are `expandedPanels` present?

jgerigmeyer: Hm, I don't think we want to disable the toggle if there are `expandedPanels` -- we just want the presence of `expandedPanels` to make it say "Hide Logs" instead of "Show Logs".

src/js/components/plans/stepsTable/index.js
```diff
@@ -54,14 +54,36 @@ class StepsTable extends React.Component<Props, State> {
...
+
+    if (panelsAreOpen && !showLogs) {
+      // empty the set
+      openPanels.clear();
+      this.setState({ expandedPanels: openPanels });
+      this.setState({ showLogs: false });
+    } else {
+      this.setState({ showLogs: !this.state.showLogs });
+    }
+  };
```
duggiemitchell: @jgerigmeyer  Def not perfect, plan to clean up tomorrow - I'm thinking of making this a case statement(?), though open to see how you would solve this.

src/js/components/plans/stepsTable/toggleLogsDataColumnLabel.js
```diff
@@ -8,20 +8,27 @@ import { t } from 'i18next';
...
+const ToggleLogsDataColumnLabel = ({
+  showLogs,
+  toggleLogs,
+  singleLog,
+}: Props) => (
   <>
     <Tooltip
       align="top left"
-      content={showLogs ? t('Hide Logs') : t('Show Logs')}
+      content={showLogs || singleLog ? t('Hide Logs') : t('Show Logs')}
```
duggiemitchell: Added this to track when there are `expandedPanels` in the parent state

src/js/components/plans/stepsTable/toggleLogsDataColumnLabel.js
```diff
@@ -8,20 +8,27 @@ import { t } from 'i18next';
 type Props = {
   showLogs: boolean,
   toggleLogs: () => void,
+  singleLog: boolean,
```
duggiemitchell: Added this to track when there are `expandedPanels` in the parent state - though maybe be named something else?

src/js/components/plans/stepsTable/index.js
```diff
@@ -44,43 +44,50 @@ type Props = {
...
-    if (panelsAreOpen && !showLogs) {
-      // empty the set
-      openPanels.clear();
-      this.setState({ expandedPanels: openPanels });
+    if (panelsAreOpen && showLogs) {
+      panels.clear();
+      this.setState({ expandedPanels: panels });
       this.setState({ showLogs: false });
     } else {
-      this.setState({ showLogs: !this.state.showLogs });
```
duggiemitchell: @jgerigmeyer I think I've figured out a cleaner way to write the branching logic in `toggleLogs` and `togglePanel`, in case you have comment on this code during review.

src/js/components/plans/stepsTable/index.js
```diff
@@ -62,32 +60,27 @@ class StepsTable extends React.Component<Props, State> {
...
+    if (steps) {
+      panelsAreOpen ? panels.clear() : steps.forEach(step => panels.add(step));
       this.setState({ expandedPanels: panels });
-      this.setState({ showLogs: false });
     } else {
-      this.setState({
-        showLogs: !this.state.showLogs,
-        autoExpand: !this.state.autoExpand,
-      });
+      return {};
```
duggiemitchell: @jgerigmeyer  the linter didn't like this giving me an error that reads: `80:7  error  Expected an assignment or function call and instead saw an expression  no-unused-expressions` When trying to commit.  tried a few different ways that also didn't pass, so just left it with this if you have a suggestion on how it should be written.

src/js/components/plans/stepsTable/nameDataCell.js
```diff
@@ -117,7 +115,7 @@ class NameDataCell extends React.Component<Props> {
                     {desc}
                   </div>
                 }
-                expanded={expandedPanels.has(id) || autoExpanded || isActive}
```
duggiemitchell: This won't work; if it's `active` users wont be able to collapse it during install.  My thought is to set an additional state of `activestep` (or something), and add that ID to the `expandedPanel` set on `<Stepstable/>`.

metadeploy/api/result_spool_logger.py
jgerigmeyer: @wlonk I think this is still open for your feedback or comment.

wlonk: Oh, I entirely missed this comment! Sorry.

duggiemitchell:
> @duggiemitchell These two (granted a bit long) tests cover the auto-expand/collapse functionality. In order to get there, they also trigger the toggle, so technically cover the `ToggleLogsDataColumnLabel` component. I think what's still missing coverage are the `togglePanel` logic (manually clicking the accordion to expand/collapse a step-log) and the `toggleLogs` logic (esp clicking to show-all and hide-all once a job is "complete").

This *really* helps.  I've written the `toggleLogs`, think its working as expected.  Will check this out and commit that and work on the `togglePanel` testing.

test/js/components/plans/stepsTable.test.js
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
-import { fireEvent, render } from 'react-testing-library';
+import { fireEvent, render, cleanup } from 'react-testing-library';
 
 import StepsTable from 'components/plans/stepsTable';
 
+afterEach(cleanup);
```
jgerigmeyer: This isn't necessary, because we're automatically running this after every test. See the following:

- https://github.com/SFDO-Tooling/MetaDeploy/blob/master/jest.config.js#L126
- https://github.com/SFDO-Tooling/MetaDeploy/blob/master/test/js/setupTests.js#L3

test/js/components/plans/stepsTable.test.js
```diff
@@ -273,6 +274,47 @@ describe('<StepsTable />', () => {
...
+            'step-1': { status: 'ok', logs: 'Test log 1' },
+            'step-2': { logs: 'Test log 2' },
+            foo: { status: 'ok', logs: 'Another test log' },
+          },
+        },
+      });
+      fireEvent.click(getByText('Step 1'));
+      const log = container.querySelector('[aria-hidden="false"] code');
+
+      expect(log).toBeInTheDocument();
```
jgerigmeyer: At this point this is just testing that there is an _expanded_ log somewhere in the list. I think you want to then click it again and test that it is collapsed?

test/js/components/plans/stepsTable.test.js
```diff
@@ -273,6 +274,47 @@ describe('<StepsTable />', () => {
...
+          'step-2': { logs: 'Test log 2' },
+          foo: { status: 'ok', logs: 'Another test log' },
+        },
+      },
+    });
+    const toggle = getAllByText('Steps')[0];
+    const steps = new Set(['step-1', 'step-2', 'step-3']);
+    fireEvent.click(toggle);
+    steps.clear();
+    expect(steps.size).toEqual(0);
```
jgerigmeyer: I'm not clear what this test is doing. `steps` isn't being used anywhere.

It should also be in a `test(..., () => { ... })` fn.

test/js/components/plans/stepsTable.test.js
duggiemitchell: > @duggiemitchell I'm not sure whether this is still WIP, but I left a few comments.

Yes it is. I got a test working that covers this step.  I realized that I didn't fire the event twice

test/js/components/plans/stepsTable.test.js
duggiemitchell: yea, I have no confidence in this 🙍🏾, but working to get the test passing for this step.

jgerigmeyer: No problem! If you'd like, let me know what your goal is for this test and I can offer some pointers.

duggiemitchell: What I was trying to do here is use some 'state', though after reading a previous comment, one I think what needs to happen is the job that we pass in should have all the steps completed, aside from that,  I can't tell what pieces of the DOM that I need to look out for for changes to know the state changed.

duggiemitchell: This is for the toggLogs logic btw 🙃

test/js/components/plans/stepsTable.test.js
```diff
@@ -289,32 +288,20 @@ describe('<StepsTable />', () => {
...
-          'step-1': { status: 'ok', logs: 'Test log 1' },
-          'step-2': { logs: 'Test log 2' },
-          foo: { status: 'ok', logs: 'Another test log' },
-        },
-      },
+      expect(log.innerHTML).toEqual('Test log 1');
+      fireEvent.click(getByText('Step 1'));
+
+      log = container.querySelector('[aria-hidden="false"] code');
+      expect(log).toBeNull();
```
duggiemitchell: @jgerigmeyer here are the changes for the `togglePanel` logic, and left a reply about the remaining test.

jgerigmeyer:
@mirisuzanne I think this is ready for merge once you sign off.

src/sass/components/_steps.scss
```diff
@@ -21,21 +59,117 @@
...
+    grid-column: name-start / step-row-end;
+    grid-row: 1 / span 2;
+  }
+
+  .plan-step-name {
+    display: flex;
+  }
+
+  .plan-step-type {
+    grid-column: type-start / type-end;
```
mirisuzanne: I think `grid-column: type;` should do it?

src/sass/components/_steps.scss
```diff
@@ -1,16 +1,54 @@
...
+
 #plan-steps-table {
   // Borders are added by parent .slds-card
   border: none;
 
+  // set up layout
+  tr {
+    display: grid;
+    grid-template-columns:
+      [step-row-start name-start] minmax($step-name-min, 4fr)
```
mirisuzanne: I think this might be simpler with areas rather than line names, so you could do both columns and rows. You'd only lose `step-row-start` and `step-row-end` - which would become `1` and `-1`, still pretty readable.

src/sass/components/_steps.scss
```diff
@@ -21,21 +59,117 @@
...
+      // center the type icon to the label only
+      span:first-child {
+        align-self: center;
+        display: flex;
+      }
+    }
+  }
+
+  .plan-step-badge-container {
+    grid-column: badge-start / badge-end;
```
mirisuzanne: Or eg `grid-area: badge` if you used areas for both row  & column.

src/sass/components/_steps.scss
```diff
@@ -21,21 +59,117 @@
...
+      margin-top: -0.3125rem;
+      margin-bottom: -0.3125rem;
+    }
+  }
+
+  section:not(.slds-is-open) > .slds-accordion__content {
+    width: 0;
+  }
+
+  .slds-is-open > .slds-accordion__content {
```
mirisuzanne: consider using custom hooks? probably not for open/closed, but for the content?

src/sass/components/_header.scss
```diff
@@ -17,6 +17,18 @@
   border: none;
   border-radius: 0;
   box-shadow: none;
+
+  nav[role='navigation'] {
+    margin-bottom: 0;
+  }
+
+  .slds-breadcrumb__item {
```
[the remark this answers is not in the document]
jgerigmeyer: Yeah, probably a good idea. This is a judgement call, and could be removed entirely. SLDS updated its spec to remove all-caps, and replaced it with bold for breadcrumbs. I added this to override because I thought the bold was too stark... but I'm happy for someone else to adjust.

src/sass/components/_header.scss
```diff
@@ -17,6 +17,18 @@
...
+    font-weight: normal;
+  }
+
+  .slds-media__figure {
+    padding-top: 0.25rem;
+  }
```
jgerigmeyer: This override is fixing what I think is a bug in react-SLDS (https://github.com/salesforce/design-system-react/issues/1972) where the vertical alignment of the icon is messed up if there's a breadcrumb.

src/sass/components/_steps.scss
stacyk: I started with grid areas but decided against it for some reason. I can't remember exactly why at the moment and am wondering if it had something to do with the pre-install table not having as many columns nor rows?

stacyk:
> Does it make sense for the logs to be inside the table-header cell? If the logs had their own cell, I think it would be simpler to lay out, and also maybe better semantics?

@jgerigmeyer is this possible?

jgerigmeyer:
> > Does it make sense for the logs to be inside the table-header cell? If the logs had their own cell, I think it would be simpler to lay out, and also maybe better semantics?
> 
> @jgerigmeyer is this possible?

I don't really understand the question/proposal. Could you show me some sample markup?

src/sass/components/_steps.scss
stacyk: @jgerigmeyer @duggiemitchell Do we have access to this slds-accordion__content to add an additional classname?

stacyk:
> > > Does it make sense for the logs to be inside the table-header cell? If the logs had their own cell, I think it would be simpler to lay out, and also maybe better semantics?
> > 
> > 
> > @jgerigmeyer is this possible?
> 
> I don't really understand the question/proposal. Could you show me some sample markup?

@mirisuzanne might have a better explanation, but I am guessing she is suggesting the log (which is in the accordion content panel) to be in it's own table cell, in the same row but at the end, perhaps, and use CSS Grid to display it on it's own row. I am guessing it isn't possible at least out of the box unless we remake their accordion pattern? Did I get that remotely close to what you were thinking Mia?

src/sass/components/_steps.scss
stacyk: check out the latest PR (in a moment). I don't know if I did what you are asking, but I think it is close.

mirisuzanne:
@stacyk @jgerigmeyer that's exactly what I meant. Can the logs have their own table cell ` in each row, rather than being part of the row-header cell?

src/sass/components/_steps.scss
jgerigmeyer: @stacyk @mirisuzanne We do not have access to the `.slds-accordion__content` element itself, no. https://github.com/salesforce/design-system-react/blob/master/components/accordion/panel.jsx#L84-L90

jgerigmeyer:
> @stacyk @jgerigmeyer that's exactly what I meant. Can the logs have their own table cell in each row, rather than being part of the row-header cell?

@mirisuzanne @stacyk Short version: no, not really. We can't separate the accordion trigger from the accordion contents, unless we remake the accordion component.

src/sass/components/_header.scss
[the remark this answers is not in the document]
jgerigmeyer: Actually... we don't have access to the markup to add a class here.

PR #265, opened 2019-03-26
[long-install] Add average time to plans

wlonk:
https://trello.com/c/UnpNfxz8/124-as-an-admin-with-a-long-running-installation-i-want-to-be-reassured-that-my-job-is-still-running

average of kittens


metadeploy/api/serializers.py
```diff
@@ -464,10 +465,17 @@ class JobSummarySerializer(serializers.ModelSerializer):
     product_slug = serializers.CharField(source="plan.version.product.slug")
     version_label = serializers.CharField(source="plan.version.label")
     plan_slug = serializers.CharField(source="plan.slug")
+    plan_average_duration = serializers.CharField(source="plan.average_duration")
```
jgerigmeyer: @wlonk This is serialized as a datetime string (e.g. `"0:00:28.589754"`), while `plan.average_duration` returns as a number of seconds string (e.g. `"28.589754"`).

src/js/utils/dates.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+  val = Math.max(Math.round((sec / 60 / 60 / 24) * 10) / 10, 1);
+  if (val < 365) {
+    msg = `${val} day${pluralize(val)}`;
+    return t('durationDays', msg, { count: val });
+  }
+  val = Math.max(Math.round(sec / 60 / 60 / 24 / 365), 1);
+  msg = `${val} year${pluralize(val)}`;
+  return t('durationYears', msg, { count: val });
+};
```
jgerigmeyer: We could get more intelligent about this (e.g. "1 hour 20 minutes" instead of "1.4 hours"), but this seemed like an okay place to start.

jgerigmeyer:
> The alert looks good. Not sure what install I need to run in order to see average install time, but I don't see anything there that worries me.

@mirisuzanne If you run an installation 5 times for the same plan, you'll start seeing the averages.

metadeploy/api/models.py
```diff
@@ -523,6 +523,18 @@ def slug_parent(self):
...
+    def average_duration(self):
+        durations = [
+            (job.success_at - job.enqueued_at)
+            for job in Job.objects.filter(
+                plan=self, status=Job.Status.complete
+            ).exclude(Q(success_at__isnull=True) | Q(enqueued_at__isnull=True))
+        ]
+        if len(durations) < settings.MINIMUM_JOBS_FOR_AVERAGE:
+            return None
+        return sum(durations, timedelta()) / len(durations)
```
davisagli: I think we should put a limit on how many jobs we include here -- partly to avoid needing to retrieve all the data when many jobs have been run, and partly so there's a rolling window to let the expected duration adjust if we suddenly add more steps that introduce a significant change in the amount of work that needs to happen.

I also suspect that it would be more appropriate to calculate the median rather than the average so that it's less susceptible to influence from outliers. If a build runs until the time limit because it got stuck somewhere, we want to minimize the effect of that on the expected build time.

locales_dev/en/translation.json
```diff
@@ -6,8 +6,11 @@
   "Access to view the installation job does not provide access to your Salesforce org.": "Access to view the installation job does not provide access to your Salesforce org.",
   "Additional Plans": "Additional Plans",
   "After resolving all errors, run the pre-install validation again.": "After resolving all errors, run the pre-install validation again.",
+  "An installation is currently running on this org.": "An installation is currently running on this org.",
   "Anyone with the link can view this installation job.": "Anyone with the link can view this installation job.",
   "Authentication Error": "Authentication Error",
+  "Average Install Time": "Average Install Time",
+  "Average install time is": "Average install time is",
```
stacyk: Since this is a sentence, would it read better to say: 
"The average install time is ..." instead of "Average install time is ..."?

metadeploy/api/models.py
jgerigmeyer: These changes make sense to me. @wlonk Can you do these in a new PR?

PR #269, opened 2019-04-01
[reconnect-data] 🐈 Flush Redux store on socket reconnect.

wlonk:
catte


src/js/index.js
```diff
@@ -122,6 +123,20 @@ init_i18n(() => {
...
+          method: 'GET',
+        }).then(user => {
+          appStore.dispatch({ type: 'USER_LOGGED_OUT' });
+          appStore.dispatch({
+            type: 'USER_LOGGED_IN',
+            user,
+          });
+          appStore.dispatch(fetchOrgJobs());
+        });
+        /* */
```
wlonk: @jgerigmeyer OK, it's this block that I'm stuck on; how do I get `apiFetch` in this context? And then, I think, there are some problems possibly with this leaving the "socket disconnected" banner up as a side-effect, so we should be alert for that.

jgerigmeyer: @wlonk Pushed some tweaks -- give it a try and see if it works as you'd expect.

jgerigmeyer:
@wlonk Once CI passes, this is good to merge as far as I'm concerned.

PR #271, opened 2019-04-02
📦 Upgrade deps to silence warnings.

jgerigmeyer:
marmot

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -6,7 +6,7 @@ channels==2.1.7
 channels_redis==2.3.3
 service_identity==18.1.0
 django-allauth==0.39.1
-werkzeug==0.15.1
+werkzeug==0.15.2
 pyopenssl==19.0.0
 django-js-reverse==0.8.2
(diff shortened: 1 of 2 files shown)


requirements/base.txt
```diff
@@ -20,13 +20,12 @@ rq==0.12  # pyup: <0.13
 django-rq==1.2.0  # pyup: <1.3.0
```
jgerigmeyer: @wlonk Could you revisit whether these pyup overrides can be removed now?

wlonk: Mm, yes, I will.

wlonk: We're stuck pinning these until (django-rq-scheduler)[https://github.com/istrategylabs/django-rq-scheduler] updates, which it hasn't in 9 months.

It lists its dependencies incorrectly, with just raw `>=` annotation, so if we allowed its deps to be resolved as it says, it'd update these packages, and `redis>=3` would break it.

If we can prioritize forking `django-rq-scheduler`, we may want to.

PR #275, opened 2019-04-08
[markdown] hides header if header in markdown

duggiemitchell:
Trello story: https://trello.com/c/1ZjDy7jl

uglydog

### DIFF
--- a/src/js/components/products/detail.js
+++ b/src/js/components/products/detail.js
@@ -134,10 +134,10 @@ class VersionDetail extends React.Component<VersionDetailProps> {
       primary_plan && primary_plan.is_listed && primary_plan.is_allowed;
     const visibleSecondaryPlan =
       secondary_plan && secondary_plan.is_listed && secondary_plan.is_allowed;
-    const productDescriptionHasTitle = product.description
-      ? product.description.startsWith('<h1>') ||
-        product.description.startsWith('<h2')
-      : null;
+    const productDescriptionHasTitle =
+      product.description &&
+      (product.description.startsWith('<h1>') ||
+        product.description.startsWith('<h2>'));
     return (
       <DocumentTitle title={`${product.title} | ${window.SITE_NAME}`}>
         <>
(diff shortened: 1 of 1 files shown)


src/js/components/products/detail.js
```diff
@@ -134,6 +134,9 @@ class VersionDetail extends React.Component<VersionDetailProps> {
       primary_plan && primary_plan.is_listed && primary_plan.is_allowed;
     const visibleSecondaryPlan =
       secondary_plan && secondary_plan.is_listed && secondary_plan.is_allowed;
+    const productDescriptionHasTitle = product.description
+      ? product.description.startsWith('<h2>')
+      : null;
```
jgerigmeyer: I think we want to expand this at least to include h1-h2?

```jsx
    const productDescriptionHasTitle = product.description
      && (product.description.startsWith('<h1>') || product.description.startsWith('<h2>');
```

duggiemitchell: OK I'll update the PR to include h2.  Are there any other areas that may be affected by this off the top?  I didn't see any other divs with the `markdown` class preceded by a header 🤷‍♀️

duggiemitchell: @jgerigmeyer this is my first time using Flow, so forgive me if the answer is obvious.  Adding the above line does not pass linting: 

```Cannot call `product.description.startsWith` because property `startsWith` is missing in null (see ./src/js/store/products/reducer.js:21)```

due to the null checks in the `Product` type.  Please advise if there is another way to define this property?

jgerigmeyer: > @jgerigmeyer this is my first time using Flow, so forgive me if the answer is obvious. Adding the above line does not pass linting:
> 
> `` Cannot call `product.description.startsWith` because property `startsWith` is missing in null (see ./src/js/store/products/reducer.js:21) ``
> 
> due to the null checks in the `Product` type. Please advise if there is another way to define this property?

Weird, that seems to be a bug in Flow -- what you have should work. I was able to fix it by repeating the `product.description &&` check, but that shouldn't be strictly necessary. 😕

jgerigmeyer: > OK I'll update the PR to include h2. Are there any other areas that may be affected by this off the top? I didn't see any other divs with the `markdown` class preceded by a header 🤷‍♀️

I looked through all the other uses of `.markdown`, and none of them follow headers like this. So I think it's good to go.

PR #278, opened 2019-04-09
[optional-steps-clarity] optional steps are highlighted yellow

duggiemitchell:
see: https://trello.com/c/1ZjDy7jl

dog2

### DIFF
--- a/src/sass/components/_steps.scss
+++ b/src/sass/components/_steps.scss
@@ -34,9 +34,7 @@
     padding-left: 0.5rem;
   }
 
-  .slds-badge_inverse {
+  .slds-badge_lightest {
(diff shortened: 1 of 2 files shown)


src/sass/components/_steps.scss
```diff
@@ -34,4 +33,10 @@
     border-left: 1rem solid $color-border-error;
     padding-left: 0.5rem;
   }
+
+  .slds-badge_inverse {
```
jgerigmeyer: If we use `.slds-badge_lightest` here (and make the corresponding change in `requiredDataCell.js`, then I think we'd only have to override `background-color` instead of all three styles?

PR #279, opened 2019-04-10
[job-url-path] 🐈 Fail to load job if product/version/plan incorrect

wlonk:
See https://trello.com/c/Xj1UoMOX/138-bug-i-can-access-a-job-detail-page-if-i-have-jobid-even-with-the-wrong-product-version-plan

catte


src/js/store/jobs/actions.js
```diff
@@ -61,13 +62,20 @@ export type JobsAction =
...
+  version: string,
+  plan: string,
+): ThunkAction => (dispatch, getState, { apiFetch }) => {
   dispatch({ type: 'FETCH_JOB_STARTED', payload: jobId });
-  return apiFetch(window.api_urls.job_detail(jobId))
+  const url = window.api_urls.job_detail(jobId);
+  const params = stringify({
+    plan__plan_template__planslug__slug: plan,
+    plan__version__label: version,
+    plan__version__product__productslug__slug: product,
```
jgerigmeyer: Magic strings? I guess you just have to know Django's secret language. :)

src/js/store/jobs/actions.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 
 import type { ThunkAction } from 'redux-thunk';
+import { stringify } from 'query-string';
```
jgerigmeyer: We have an existing `addUrlParams` helper we can use here.

src/js/components/jobs/detail.js
```diff
@@ -81,10 +81,15 @@ class JobDetail extends React.Component<Props, State> {
...
+    const { job, jobId, doFetchJob, product, versionLabel, plan } = this.props;
     if (job === undefined && jobId) {
       // Fetch job from API
-      doFetchJob(jobId);
+      doFetchJob(
+        jobId,
+        product ? product.slug : '',
+        versionLabel || '',
+        plan ? plan.slug : '',
+      );
```
jgerigmeyer: This fixes the bug for page-load fetching of a `job`, but I just realized it still doesn't do a check after we have a fetched job. That is, once we know about a job, when you try to navigate to that job detail page there's nowhere we check to be sure the job you're viewing matches the product/version/plan you're viewing.

src/js/components/jobs/detail.js
```diff
@@ -81,10 +81,15 @@ class JobDetail extends React.Component<Props, State> {
   }
 
   fetchJobIfMissing() {
-    const { job, jobId, doFetchJob } = this.props;
-    if (job === undefined && jobId) {
+    const { job, jobId, doFetchJob, product, versionLabel, plan } = this.props;
+    if (product && versionLabel && plan && jobId && job === undefined) {
```
jgerigmeyer: No point in fetching the `job` if we don't have the params we need.

src/js/store/jobs/selectors.js
```diff
@@ -14,13 +16,18 @@ export const selectJobId = (
...
       return undefined;
     }
     // A `null` job means we already fetched and no prior job exists
     // An `undefined` job means we don't know whether a job exists
-    return jobs[jobId];
+    const job = jobs[jobId];
+    // Only return job if it belongs to the requested plan...
+    if (!job || (plan && job.plan === plan.id)) {
+      return job;
+    }
```
jgerigmeyer: This should fix bug #2 I mentioned.

src/js/components/plans/detail.js
```diff
@@ -190,6 +190,7 @@ class PlanDetail extends React.Component<Props, State> {
     }
     if (org) {
       if (org.current_job) {
+        const { product_slug, version_label, plan_slug, id } = org.current_job;
```
jgerigmeyer: ...and this fixed bug #3.

PR #282, opened 2019-04-11
[allowed-list-org-type] 🛢Allow all orgs that match a type through an AllowedList

wlonk:
See https://trello.com/c/UAVZHpy6/133-as-sf-staff-i-want-to-add-an-org-type-to-an-allowedlist

cat-puter


metadeploy/api/models.py
```diff
@@ -54,8 +54,17 @@ class Meta:
...
+        ("", ""),
+        ("Production", "Production"),
+        ("Scratch", "Scratch"),
+        ("Sandbox", "Sandbox"),
+        ("Developer", "Developer"),
+    )
+
     title = models.CharField(max_length=128, unique=True)
     description = MarkdownField(blank=True, property_suffix="_markdown")
+    org_type = models.CharField(max_length=64, choices=ORG_TYPES, blank=True)
```
davisagli: We need to be able to select multiple allowed org types for a single AllowedList.

An ArrayField is probably fine as opposed to a separate model.

Can we add a description in the admin along the lines of "All orgs of these types will be automatically allowed."

wlonk:
See commit message in 86974c9:

```
THIS CHANGES A MIGRATION THAT WAS IN THIS BRANCH.

Before applying this commit, run::

    python manage.py migrate api 0064

Then once you've applied this commit, run::

    python manage.py migrate

If you miss this instruction, you can run::

    git checkout c29403b
    python manage.py migrate api 0064
    git checkout allowed-list-org-type
```

jgerigmeyer:
@wlonk In the admin I only see a text field -- does the user have to know the exact terms to enter?

It'd be nice to have a multiselect or checkboxes or something. Also missing help text.

wlonk:
Yeah, I'm trying to figure out how to make a multi-select out of something other than an FK. The Django admin doesn't handle ArrayFields well, imo. Fixes incoming when I work them out.

wlonk:
Wrangled the admin interface! It's good now.

PR #287, opened 2019-04-12
[error-clarity] Make installation errors more obvious (and helpful).

jgerigmeyer:
https://trello.com/c/0WVmmFUs

red panda


jgerigmeyer:
Needs js-test coverage and adding `job.error_message` to the API. Functionally, here are the changes:

Failed installation:

Screenshot 2019-04-12 16 38 57

Failed installation modal (opens automatically when job fails, but not on every subsequent page-load):

Screenshot 2019-04-12 16 39 49

Modal with fallback error message (if neither plan nor product has one):

Screenshot 2019-04-12 16 39 05

jgerigmeyer:
@davisagli @wlonk @SondraE I think this is ready for merge pending code and functional review.

locales_dev/en/translation.json
```diff
@@ -103,14 +109,14 @@
   "pageCannotBeFound": "That page cannot be found. Try the <2>home page</2>?",
   "planNotAllowed": "Try <2>another plan</2> from that product version",
   "planNotFound": "We can’t find the plan you’re looking for. Try <2>another plan</2> from that product version?",
-  "preflightValidTime": "Pre-install validation will expire if install is not run within <1>{{preflight_minutes}}</1> minutes, and you will need to run pre-install validation again.",
-  "preflightValidTime_plural": "Pre-install validation will expire if install is not run within <1>{{preflight_minutes}}</1> minutes, and you will need to run pre-install validation again.",
+  "preflightValidTime": "Pre-install validation will expire if install is not run within <2>{{preflight_minutes}}</2> minutes, and you will need to run pre-install validation again.",
+  "preflightValidTime_plural": "Pre-install validation will expire if install is not run within <2>{{preflight_minutes}}</2> minutes, and you will need to run pre-install validation again.",
```
wlonk: Why is this `<2>` now?

metadeploy/api/models.py
```diff
@@ -318,6 +319,10 @@ def description_markdown(self):
     def click_through_agreement_markdown(self):
         return self.get_translation("en-us").click_through_agreement_markdown
 
+    @property
+    def error_message_markdown(self):
```
wlonk: I know this is part of what I committed, but I just wanna say I don't like this "add properties that get the rendered translation" and would like to maybe metaprogram my way out of it.

locales_dev/en/translation.json
jgerigmeyer: I'm not entirely sure, but i18next uses these as auto-generated placeholders for other elements. Perhaps it counts the `{' '}` as something requiring a placeholder?

davisagli:
Sometime in the future we will have a database of errors and the ability to provide custom text based on which error was encountered. We don't have that yet, so this is an intermediate step to provide some general direction when errors happen configured at the plan template or product level. Because it's general, we do need metadeploy to take care of also showing the specific step and error message in the modal.

jgerigmeyer:
Updated fallback message, and now showing step-specific error in the modal.

Screenshot 2019-04-18 15 34 05

PR #294, opened 2019-04-18
[debounce-websockets] Debounce websockets

wlonk:
This just sets a semaphore in Redis where the message is set, and pulls it where the message is read.

See https://trello.com/c/2abLmB1E/140-consider-debouncing-throttling-websocket-events

cat

### DIFF
--- a/metadeploy/api/tests/push.py
+++ b/metadeploy/api/tests/push.py
@@ -24,12 +24,12 @@ async def test_report_error(mocker, user_factory):
 @pytest.mark.django_db
 @pytest.mark.asyncio
 async def test_notify_org_job_changed(mocker, user_factory, job_factory, plan_factory):
-    get_channel_layer = mocker.patch("metadeploy.api.push.get_channel_layer")
-    channel_layer = MagicMock(name="channel_layer")
-    channel_layer.group_send = AsyncMock(name="group_send")
-    get_channel_layer.return_value = channel_layer
+    # get_channel_layer = mocker.patch("metadeploy.api.push.get_channel_layer")
+    # channel_layer = MagicMock(name="channel_layer")
+    # channel_layer.group_send = AsyncMock(name="group_send")
+    # get_channel_layer.return_value = channel_layer
     user = user_factory()
     plan = plan_factory()
     job = job_factory(user=user, plan=plan, organization_url="https://example.com/")
     await notify_org_result_changed(job)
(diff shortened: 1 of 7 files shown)


wlonk:
@jgerigmeyer do you wanna take a look at this and lemme know if it seems to work for you, before I figure out how to get the tests working? (Which will require some mocking out of Redis.)

jgerigmeyer:
@wlonk This does seem to be an improvement. I'm now getting all the events I'd expect, and fewer events with duplicate payloads. Before, I was getting ~108 `TASK_COMPLETED` events with lots of duplicates. With this PR, I'm getting ~69, with only a handful of duplicates (but there are still some). I think it's still more frequent log updates than we'd need, but 🤷‍♂️

metadeploy/api/tests/push.py
```diff
@@ -24,12 +24,12 @@ class AsyncMock(MagicMock):
...
+    # get_channel_layer = mocker.patch("metadeploy.api.push.get_channel_layer")
+    # channel_layer = MagicMock(name="channel_layer")
+    # channel_layer.group_send = AsyncMock(name="group_send")
+    # get_channel_layer.return_value = channel_layer
     user = user_factory()
     plan = plan_factory()
     job = job_factory(user=user, plan=plan, organization_url="https://example.com/")
     await notify_org_result_changed(job)
-    assert channel_layer.group_send.called
+    # assert channel_layer.group_send.called
```
jgerigmeyer: Remove these commented lines?

wlonk: Oh, no—I need to actually test something there. So, I'll fix that line and uncomment it.

davisagli:
Hmm. There's still a race condition here, right? Because if two coroutines get past the `if not message_exists()` check before either one gets to set_message_exists(), then they will both set the semaphore and send the message.

To avoid this, could you use the NX option to redis SET so that there's a single call to redis that only sets the semaphore if it's not already set, and use the result to determine whether to send the message? https://redis.io/commands/SET

wlonk:
@davisagli Thanks; I don't know much about Redis, but this looks like it'll be a step towards what we really need.

jgerigmeyer:
Functionally this gives me the same results as before: ~68 events, with a handful of identical duplicates.

jgerigmeyer:
@davisagli I think this is ready for merge once you've signed off on it.

PR #305, opened 2019-04-29
[sfdo-t-h-update] 📦 Update version of SFDO Template Helpers

wlonk:
Just a quick requirements update.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
             . venv/bin/activate
             # @@@ temp fix
             rm -rf ./venv/src/sfdo-template-helpers
-            pip install -r requirements/local.txt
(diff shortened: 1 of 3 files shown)


davisagli:
@wlonk I'm not sure why it isn't using the newer version of cryptography that is in cumulusci's requirements, but I bet adding it explicitly to metadeploy's requirements would help

wlonk:
@davisagli yeah, I'd like to not do that if I can! I'm gonna try one more thing.

.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ jobs:
             . venv/bin/activate
             # @@@ temp fix
             rm -rf ./venv/src/sfdo-template-helpers
```
jgerigmeyer: We should be able to remove this line now?

PR #307, opened 2019-04-30
[ansi-colors] Colorize logs

wlonk:
See https://trello.com/c/4BvhuzaZ/141-as-an-admin-i-want-to-see-ansi-colors-in-my-installation-step-logs

sloth baby at the Denver zoo


jgerigmeyer:
@wlonk I'm not seeing any changes here functionally (even when I run an installation of Education Data Architecture). What do I need to do to test this?

wlonk:
@jgerigmeyer I'm not sure, then. I would have expected that would do it, based on the sample usage of it we were sent. I'll dig into this more; it's supposed, as far as I know, to colorize things based on the kwargs passed in?

jgerigmeyer:
> @jgerigmeyer I'm not sure, then. I would have expected that would do it, based on the sample usage of it we were sent. I'll dig into this more; it's supposed, as far as I know, to colorize things based on the kwargs passed in?

I should clarify that since the front-end is not using `dangerouslySetInnerHTML` (yet), I would not expect this to just work. But the logs I'm seeing are still strings without HTML, so.

wlonk:
@jgerigmeyer Right, that makes sense.

wlonk:
The code as-written correctly colorizes things in the abstract, but the data passed to the logger doesn't have any fields except the `msg`, so it doesn't colorize it. Not sure what to do there.

metadeploy/api/result_spool_logger.py
```diff
@@ -10,18 +12,21 @@ def __init__(self, *args, result=None, **kwargs):
     def emit(self, record):
         if self.current_key is None:
             return
+        msg = record.getMessage()
```
davisagli: @wlonk I think this needs to be `msg = self.format(record)` so that you get the message as processed by the formatter instead of the raw message that was logged.

wlonk:
@davisagli caught it, now we should be good to go with the frontend components.

jgerigmeyer:
LGTM!

@mirisuzanne @SondraE @stacyk Any preference for which color scheme we use? All of them are pretty similar... see https://github.com/ralphbean/ansi2html/blob/master/ansi2html/style.py#L49-L80

default:
Screenshot 2019-05-23 18 20 41

solarized:
Screenshot 2019-05-23 18 17 33

xterm:
Screenshot 2019-05-23 18 22 56

osx:
Screenshot 2019-05-23 18 25 15

mint-terminal:
Screenshot 2019-05-23 18 26 31

mirisuzanne:
I'm happy with any of those, if @stacyk or @SondraE has a preference

jgerigmeyer:
I have a slight leaning for "solarized". ¯\_(ツ)_/¯

stacyk:
I think osx is the easiest to read... but barely.

PR #310, opened 2019-05-03
[documentation] 📦 Add documentation stubs

wlonk:
See https://trello.com/c/3bAkUkDP/84-as-salesforce-staff-i-want-api-documentation

rhino


docs/conf.py
```diff
@@ -30,7 +30,11 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ["sphinx.ext.autodoc"]
+extensions = [
+    "sphinx.ext.autodoc",
+    "sphinxcontrib.httpdomain",
+    "sphinxcontrib.httpexample",
```
davisagli: Does this one need to be added to the requirements also, or is it not actually used?

wlonk: Not actually used! Left over from earlier experimentation.

wlonk:
@davisagli Ready for another review if you like!

PR #313, opened 2019-05-08
[fetch-plans] Fetch plans on-demand.

jgerigmeyer:
https://trello.com/c/Z8uWlkKD

tiger


src/js/store/products/reducer.js
```diff
@@ -56,6 +56,27 @@ const reducer = (products: Products = [], action: ProductsAction): Products => {
...
+            fetched_additional_plans: true,
+          };
+          return {
+            ...p,
+            versions,
+          };
+        }
+        return p;
+      });
+    }
```
duggiemitchell: @jgerigmeyer a couple things about this:
1.  are the checks on line 66 the right ones?
2.  it seems to be adding the object to the state store, but seems to keep calling the `FETCH_PLANS_?` actions (according to the console), not sure how to tell why 🤷🏽.

src/js/store/products/actions.js
```diff
@@ -3,7 +3,7 @@
 import type { ThunkAction } from 'redux-thunk';
 
 import { addUrlParams } from 'utils/api';
-import type { Products, Version } from 'store/products/reducer';
+import type { Product, Products, Version } from 'store/products/reducer';
```
jgerigmeyer: I don't see this import being used.

src/js/store/products/actions.js
```diff
@@ -26,7 +26,7 @@ type FetchVersionFailed = {
 };
 type FetchPlansStarted = {
   type: 'FETCH_PLANS_STARTED',
-  payload: [], // think this needs to be a type like Version.additional_plans ?
+  payload: string,
```
jgerigmeyer: For each of these, I think the payload might as well be `{ product: string, version: string }`, same as the argument passed into the action.

src/js/store/products/actions.js
```diff
@@ -98,7 +98,7 @@ export const fetchVersion = (filters: VersionFilters): ThunkAction => (
     });
 };
 
-export const fetchPlans = (version: string): ThunkAction => (
+export const fetchPlans = (product: string, version: string): ThunkAction => (
```
jgerigmeyer: For consistency with other actions, let's combine these into one payload object: `{ product: string, version: string }`

src/js/store/products/actions.js
```diff
@@ -116,7 +116,7 @@ export const fetchPlans = (version: string): ThunkAction => (
       }
       return dispatch({
         type: 'FETCH_PLANS_SUCCEEDED',
-        payload: response,
+        payload: { response, product, version },
```
jgerigmeyer: 👍

src/js/store/products/reducer.js
```diff
@@ -56,6 +56,27 @@ const reducer = (products: Products = [], action: ProductsAction): Products => {
         return p;
       });
     }
+    case 'FETCH_PLANS_SUCCEEDED': {
+      const { response, product, version } = action.payload;
```
jgerigmeyer: This causes Flow errors because the action type isn't defined properly in `actions.js`:

```js
type FetchPlansStarted = {
  type: 'FETCH_PLANS_STARTED',
  payload: { product: string, version: string },
};
type FetchPlansFailed = {
  type: 'FETCH_PLANS_FAILED',
  payload: { product: string, version: string },
};
type FetchPlansSucceeded = {
  type: 'FETCH_PLANS_SUCCEEDED',
  payload: { product: string, version: string, response: Array<Plan> },
};
```

src/js/store/products/reducer.js
```diff
@@ -56,6 +56,27 @@ const reducer = (products: Products = [], action: ProductsAction): Products => {
...
+      const additional_plans = response.reduce((obj, item) => {
+        obj[item.id] = item;
```
jgerigmeyer: This should be `obj[item.slug]` for ease of lookup, I think.

Okay, so to confuse things... each `plan` also has a list of `old_slugs`. I think it will simplify things elsewhere if we store each plan on the `obj` both under `item.slug` and under _each_ entry in `item.old_slugs`. Maybe e.g.:

```js
      const additional_plans = response.reduce((obj, item) => {
        obj[item.slug] = item;
        for (const oldSlug of item.old_slugs) {
          obj[oldSlug] = item;
        }
        return obj;
      }, {});
```

src/js/store/products/reducer.js
```diff
@@ -56,6 +56,27 @@ const reducer = (products: Products = [], action: ProductsAction): Products => {
...
+          const versions = {
+            ...p.versions,
+            additional_plans,
+            fetched_additional_plans: true,
+          };
+          return {
+            ...p,
+            versions,
+          };
+        }
```
jgerigmeyer:
```js
        if (p.id === product) {
          if (p.most_recent_version && p.most_recent_version.id === version) {
            return {
              ...p,
              most_recent_version: {
                ...p.most_recent_version,
                fetched_additional_plans: true,
                additional_plans,
              },
            };
          } else if (p.versions) {
            const thisVersion: ?Version = (Object.values(p.versions): any).find(
              (v: Version | null) => v !== null && v.id === version,
            );
            if (thisVersion) {
              return {
                ...p,
                versions: {
                  ...p.versions,
                  [thisVersion.label]: {
                    ...thisVersion,
                    fetched_additional_plans: true,
                    additional_plans,
                  },
                },
              };
            }
          }
        }
```

src/js/store/products/reducer.js
jgerigmeyer: I think my proposed changes hopefully address all three?

src/js/components/utils.js
```diff
@@ -183,6 +183,7 @@ export const getLoadingOrNotFound = ({
     if (!version) {
       return <VersionNotFound product={product} />;
     }
+    // @@@ I'm not sure this is the right condition we are looking for
     if (
       !version.additional_plans ||
       (planSlug && version.additional_plans[planSlug] === null)
```
duggiemitchell: @jgerigmeyer I don't think this block is not read on execution, can you confirm the condition we need to check in order to display the spinner while fetching plans?

src/js/store/products/reducer.js
```diff
@@ -55,6 +56,48 @@ const reducer = (products: Products = [], action: ProductsAction): Products => {
...
+                    ...thisVersion,
+                    fetched_additional_plans: true,
+                    additional_plans,
+                  },
+                },
+              };
+            }
+          }
+        }
+        return p;
```
jgerigmeyer: @duggiemitchell This missing `return` was causing the regression you noticed 👍

src/js/components/plans/detail.js
```diff
@@ -94,6 +95,28 @@ class PlanDetail extends React.Component<Props, State> {
...
+    if (!plan) {
+      if (
+        product.id !== primary_plan.id ||
+        (secondary_plan.id && version.additional_plans === undefined)
+      ) {
+        doFetchPlan(versionLabel, productSlug, planSlug);
+      }
+    }
+  }
+
```
duggiemitchell: @jgerigmeyer or @wlonk this function works, but continue to get flow error about calling properties on possibly null values.  I can't seem to find the trick to get around these types of errors, and think it's also causing a test to break. 🤷🏽

wlonk: I can take a look tonight then.

jgerigmeyer: It seems like there's some confusing between products/versions/plans in this function. I think the goal is something like:

1. Do we have the plan from `version.primary_plan` or `version.secondary_plan`? If yes, show it. If no...
2. Does the plan exist in `version.additional_plans`? If yes, show it. If no...
3. Is it `undefined` or `null`? If `undefined`, fetch it. If `null`, show a 404.

#1 and #2 should be handled by the selector. The logic for #3 is handled by `shouldFetchPlan` in `components/utils.js`. So I think all this needs to do is combine/call some of those. E.g.

```js
    const { plan, version, productSlug, versionLabel, planSlug, doFetchPlan } = this.props;
    if (
      !plan &&
      version &&
      productSlug &&
      versionLabel &&
      planSlug &&
      shouldFetchPlan({ version, planSlug })
    ) {
      // Fetch plan from API
      doFetchPlan(versionLabel, productSlug, planSlug);
    }
```

duggiemitchell:
@jgerigmeyer @wlonk running `yarn test` should give you the jest (excuse the pun) of the current state of front end tests.  I can also leave notes about specific things that I couldn't figure out.  

Also, there may be one requirement for this story I think that's missing which is: adding a plan as null when fetched and displaying a 404.

src/js/store/plans/actions.js
```diff
@@ -50,6 +51,36 @@ export type PlansAction =
...
+export const fetchPlan = (
+  versionLabel: string,
+  productSlug: string,
+  planSlug: string,
+): ThunkAction => (dispatch, getState, { apiFetch }) => {
+  dispatch({ type: 'FETCH_PLAN_STARTED', payload: {} });
+  const baseUrl = window.api_urls.plan_list();
+  const params = { versionLabel, productSlug, planSlug };
+  return apiFetch(
+    // @todo pass into fn, had unexpected behavior using the addUrl function
```
jgerigmeyer: @duggiemitchell I'm curious what unexpected behavior you were seeing here?

duggiemitchell: @jgerigmeyer wish I'd written down exactly what happened. Basically the query string wasn't created as I expected it would. I think instead of the url reading `versional_label=` it was `Object%` instead.  
It did return a response, just not the one we needed.

metadeploy/api/filters.py
```diff
@@ -0,0 +1,12 @@
...
+from .models import Plan
+
+
+class PlanFilter(filters.FilterSet):
+    slug = filters.CharFilter(field_name="plan_template__planslug__slug")
+    product = filters.CharFilter(field_name="version__product")
+
+    class Meta:
+        model = Plan
+        fields = ("slug", "version", "product")
```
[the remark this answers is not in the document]
wlonk: That looks right, yep.

src/js/components/products/detail.js
```diff
@@ -97,20 +100,47 @@ class VersionDetail extends React.Component<VersionDetailProps> {
...
+        version: maybeVersion,
+        slug: maybeSlug,
+      });
+    }
+  }
+
   componentDidMount() {
     this.fetchVersionIfMissing();
+    this.fetchPlansIfMissing();
+    this.fetchPlanIfMissing();
```
wlonk: I feel confident these names will be confusing, and get confused. Can we rename them to differ more?

src/js/components/products/detail.js
```diff
@@ -140,9 +172,14 @@ class VersionDetail extends React.Component<VersionDetailProps> {
...
-    const listedAdditionalPlans = version.additional_plans.filter(
-      plan => plan.is_listed && plan.is_allowed,
-    );
+    const listedAdditionalPlans: Array<PlanType> = version.additional_plans
+      ? (Object.entries(version.additional_plans): any)
+          .filter(
+            ([key, plan]: [string, PlanType | null]) =>
+              plan && plan.is_listed && plan.is_allowed && key === plan.slug,
+          )
+          .map((item: Array<[string, PlanType]>) => item[1])
```
wlonk: Is all of the explicit type annotation necessary here? If so, that's OK, just wanna check.

test/js/components/utils.test.js
```diff
@@ -240,60 +311,30 @@ describe('getLoadingOrNotFound', () => {
...
+  //       version: {
+  //         id: 'v1',
+  //         additional_plans: {
+  //           'old-plan': null,
+  //         },
+  //       },
+  //     });
+  //     expect(getByText('Loading...')).toBeVisible();
+  //   });
+  // });
```
duggiemitchell: @jgerigmeyer did you intend to keep this commented out?

test/js/setupTests.js
```diff
@@ -6,17 +6,18 @@ import fetchMock from 'fetch-mock';
...
+    plan_list: () => '/api/plans/',
+    plan_preflight: id => `/api/plans/${id}/preflight/`,
+    product_list: () => '/api/products/',
+    salesforce_custom_login: () => '/accounts/salesforce-custom/login/',
+    salesforce_production_login: () => '/accounts/salesforce-production/login/',
+    salesforce_test_login: () => '/accounts/salesforce-test/login/',
     user: () => '/api/user/',
-    plan_list: () => 'api/plans/',
+    version_additional_plans: id => `/api/versions/${id}/additional_plans/`,
+    version_list: () => '/api/versions/',
```
duggiemitchell: Could we use the sort-keys rule to enforce alphabetical order? https://eslint.org/docs/rules/sort-keys

src/js/store/products/selectors.js
```diff
@@ -116,12 +116,15 @@ const selectVersionLabelOrPlanSlug: (
...
       const slugs = [];
+      /* istanbul ignore else */
       if (most_recent_version.primary_plan) {
         slugs.push(most_recent_version.primary_plan.slug);
       }
+      /* istanbul ignore else */
       if (most_recent_version.secondary_plan) {
         slugs.push(most_recent_version.secondary_plan.slug);
       }
+      /* istanbul ignore else */
```
duggiemitchell: @jgerigmeyer what do you use to determine whether to have the tests ignore if/else branches?

jest.config.js
```diff
@@ -50,7 +50,12 @@ module.exports = {
   // errorOnDeprecated: false,
 
   // Force coverage collection from ignored files usin a array of glob patterns
-  // forceCoverageMatch: [],
+  forceCoverageMatch: [
+    'src/js/**/*.{js,jsx}',
+    '!src/js/index.js',
+    '!src/js/raven.js',
+    '!src/js/i18n.js',
+  ],
```
duggiemitchell: I noticed that sometimes, the coverage report would omit some files (when you look at them from the directory, but you can find it if typing in the url).  Does this fix that?

test/js/components/utils.test.js
duggiemitchell: Disregard,looks like this is outdated.

src/js/components/products/detail.js
jgerigmeyer: Yes and no. Flow doesn't work well with `Object.entries()` (or `.keys` or `.values`), so if we want any type-checking here at all we need to be explicit. Maybe it's too much, but I also appreciate seeing the explicit changes with chained fns like this.

test/js/setupTests.js
jgerigmeyer: Yeah, maybe that's a good idea. I'm sort of inconsistent when it comes to this -- sometimes I prefer grouping in ways that seem logical to me (that is, `id` comes first), while other times I sort alphabetically. 🤷‍♂ Probably a good idea to standardize on the next project?

src/js/store/products/selectors.js
jgerigmeyer: That's a great question. I try not to overuse it, but I feel like it's fine skipping coverage on a branch that's overly-defensive -- that is, it's impossible to reproduce in practice, and/or doesn't dramatically alter the logic of the fn.

jest.config.js
jgerigmeyer: I thought it might, but it didn't really (so I reverted this change later). I think the root cause is this: https://github.com/facebook/jest/issues/7331

PR #314, opened 2019-05-08
[improve-average-duration] Use median for "average duration"

wlonk:
See https://github.com/SFDO-Tooling/MetaDeploy/pull/265#pullrequestreview-234704309

klipspringer

### DIFF
--- a/metadeploy/api/models.py
+++ b/metadeploy/api/models.py
@@ -1,6 +1,7 @@
 import itertools
 import logging
 from datetime import timedelta
+from statistics import median
 from typing import Union
 
 from allauth.socialaccount.models import SocialToken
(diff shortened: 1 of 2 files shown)


metadeploy/api/models.py
```diff
@@ -527,13 +528,13 @@ def slug_queryset(self):
     def average_duration(self):
         durations = [
             (job.success_at - job.enqueued_at)
-            for job in Job.objects.filter(
-                plan=self, status=Job.Status.complete
-            ).exclude(Q(success_at__isnull=True) | Q(enqueued_at__isnull=True))
+            for job in Job.objects.filter(plan=self, status=Job.Status.complete)
+            .exclude(Q(success_at__isnull=True) | Q(enqueued_at__isnull=True))
+            .order_by("-created_at")[:20]
```
wlonk: @davisagli I just set the max number of entries to an arbitrary value; I'd like to pull it into settings like on like 535 below, but I'd like to discuss what it should be first.

davisagli: I'm not sure what it should be either. Let's make sure there's an environment variable controlling it and then we can play with it in production.

metadeploy/api/models.py
```diff
@@ -1,6 +1,7 @@
 import itertools
 import logging
 from datetime import timedelta
+from statistics import median
```
davisagli: Thing I learned!

PR #315, opened 2019-05-08
[realistic-plan] Add EDA to sample data

wlonk:
THIS RUNS FOR REAL. Be warned.

zeebree

### DIFF
--- a/metadeploy/api/management/commands/populate_data.py
+++ b/metadeploy/api/management/commands/populate_data.py
@@ -13,6 +13,7 @@ from ...models import (
     PlanTemplate,
     Product,
     ProductCategory,
+    ProductSlug,
     Step,
     Version,
 )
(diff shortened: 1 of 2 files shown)


jgerigmeyer:
@wlonk Um, just to clarify... what does "THIS RUNS FOR REAL. Be warned." mean in practical terms? Should we not run this install in local dev? I thought the purpose of this was adding a real install that would still be harmless.

wlonk:
It means it will modify your Salesforce instance. I am not sure how to make it "realistic" without doing that.

jgerigmeyer:
@wlonk Right... but does that matter practically for our purposes?

wlonk:
@jgerigmeyer to be honest, I don't know

davisagli:
@jgerigmeyer @wlonk it should be fine as long as you're using a scratch org or a Developer Edition org or a sandbox rather than a production org. It should also be fine if you're using a production org and want to use EDA ;-)

PR #316, opened 2019-05-09
[hotfix-api-error] Anticipate possible `null` API response.

jgerigmeyer:
https://oddbird.slack.com/archives/CB2576XT4/p1557426417013800

armadillo

### DIFF
--- a/src/js/store/jobs/actions.js
+++ b/src/js/store/jobs/actions.js
@@ -114,8 +114,11 @@ export const startJob = (data: JobData): ThunkAction => (
     },
   })
     .then(response => {
+      if (!response) {
+        throw new Error('Invalid response received');
+      }
       /* istanbul ignore else */
-      if (response && window.socket) {
+      if (window.socket) {
         window.socket.subscribe({
           model: 'job',
           id: response.id,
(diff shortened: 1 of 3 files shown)


src/js/store/jobs/actions.js
```diff
@@ -115,7 +115,7 @@ export const startJob = (data: JobData): ThunkAction => (
   })
     .then(response => {
       /* istanbul ignore else */
-      if (window.socket) {
+      if (response && window.socket) {
```
davisagli: If we got this far we're going to dispatch JOB_STARTED which is not actually the case in the scenario we've been talking about, right?

jgerigmeyer: That is a very good point.

src/js/store/jobs/actions.js
```diff
@@ -114,8 +114,11 @@ export const startJob = (data: JobData): ThunkAction => (
     },
   })
     .then(response => {
+      if (!response) {
+        throw new Error('Invalid response received');
+      }
```
davisagli: will this be caught and shown to the user? just go into the console?

jgerigmeyer: Just the console -- at least until https://trello.com/c/DmfdVJCB

PR #317, opened 2019-05-09
[allowedlist-speakeasy] 🛢 Show is_allowed_by_org to frontend

wlonk:
Tell the frontend if a plan or product is only visible because of the user's `org_type`.

See https://trello.com/c/9D50UD3p/148-add-option-to-allowedlist-to-not-list-plan-for-orgs-allowed-by-org-type

manateeeeee


wlonk:
@jgerigmeyer I didn't filter the list, because I wanted to keep things consistent with how we did it for existing `is_allowed` values. I just added a related boolean that indicates if it's only allowed because of the user's `org_type`. So if that's the case, we should omit it from listings. Sound good?

metadeploy/api/models.py
```diff
@@ -71,6 +71,14 @@ class AllowedList(models.Model):
         default=list,
         help_text="All orgs of these types will be automatically allowed.",
     )
+    list_for_allowed_by_orgs = models.BooleanField(
```
jgerigmeyer: @wlonk I don't see this new Boolean being used anywhere. What am I missing?

wlonk: You're not! I am. I had a braino, and wrote the rest of the code assuming it's `True`.

jgerigmeyer:
> @jgerigmeyer I didn't filter the list, because I wanted to keep things consistent with how we did it for existing `is_allowed` values. I just added a related boolean that indicates if it's only allowed because of the user's `org_type`. So if that's the case, we should omit it from listings. Sound good?

@wlonk So... basically, `is_allowed_by_org: True` means the same thing for the front-end as `is_listed: False`?

wlonk:
@jgerigmeyer Yeaaaaaaaah, I could just combine those.

PR #318, opened 2019-05-09
🛢  Use org id to look up orgs, not organization_url / instance url

davisagli:
(the latter can be repeated for orgs not using custom domains)

The central change here is adding an org_id field to Job and PreflightResult so that we can query using it instead of organization_url. There's also a change to the identifier for the websocket channel for org-related messages.

@wlonk @jgerigmeyer I've tested this locally to make sure that I can still start a job and that it still successfully reports running jobs, but I would appreciate your review.


metadeploy/api/migrations/0073_org_id.py
```diff
@@ -0,0 +1,51 @@
...
+    for sa in SocialAccount.objects.all():
+        user_id_to_org_id[str(sa.user_id)] = sa.extra_data["organization_details"]["Id"]
+
+    Job = apps.get_model("api", "Job")
+    for job in Job.objects.all():
+        user_id = str(job.user_id)
+        org_id = user_id_to_org_id.get(user_id)
+        if org_id:
+            job.org_id = org_id
+            job.save()
```
wlonk: This might not matter much in a one-off migration, but you could write this to be quicker by not saving in a loop; we can talk about how if you care.

davisagli: @wlonk I was vaguely aware of that, but not especially concerned about performance given the current amount of data. I'd like to see how that works though, if there's an example you can point me to.

wlonk: Sure. In a simplex case, you use `.update` and `F(...)` statements. In a more complex case, I use stupid `Case` tricks, which I think you've seen me use before? Basically, build a list of `When` objects, which basically function as the mapping you have there, then `.update` with `Case(*whens)`.

PR #319, opened 2019-05-09
🛢 Fix migration if extra_data is missing organization_details

davisagli:


metadeploy/api/migrations/0073_org_id.py
```diff
@@ -9,7 +9,10 @@ def forwards(apps, schema_editor):
     SocialAccount = apps.get_model("socialaccount", "SocialAccount")
     user_id_to_org_id = {}
     for sa in SocialAccount.objects.all():
-        user_id_to_org_id[str(sa.user_id)] = sa.extra_data["organization_details"]["Id"]
+        org_details = sa.extra_data.get("organization_details") or {}
+        org_id = org_details.get("Id")
```
wlonk:
```suggestion
        org_id = sa.extra_data.get("organization_details", {}).get("Id")
```

davisagli: @wlonk I saw one case where organization_details was present but set to None, so wanted to handle that one too.

PR #320, opened 2019-05-10
Add basic admin docs

davisagli:


docs/tutorial.rst
```diff
@@ -0,0 +1,94 @@
...
+    plans:
+        install:
+            slug: install
+            title: Install
+            tier: primary
+            steps:
+                1:
+                    flow: install_prod
+
+This sets us up to publish the ``install_prod`` flow to a Plan that will be the primary (default) Plan for the Product and available with /install in the URL (from the slug).
```
wlonk:
```suggestion
This sets us up to publish the ``install_prod`` flow to a Plan that will be the primary (default) Plan for the Product and available with ``/install`` in the URL (from the slug).
```

docs/tutorial.rst
```diff
@@ -0,0 +1,94 @@
...
+
+4. Connect CumulusCI to MetaDeploy
+==================================
+
+This is a one-time setup task.
+In the MetaDeploy admin UI, go to the Tokens object (under Auth Tokens) and create one for your user.
```
wlonk: You can also use the `manage` command, if you have access to that. Should that be included here, or does this presume only web access?

davisagli: I'm fine mentioning it, but the web-based version is easier to explain given it's what we're describing above.

wlonk: Then I have no attachment to mentioning it.

PR #321, opened 2019-05-13
🛢 Better default for list_for_allowed_by_orgs

davisagli:
Make it private by default.

(I'm intentionally modifying the existing migration rather than creating a new one because it hasn't been run on production yet.)


davisagli:
@wlonk that's a good instinct in general, but in this case we're adding the new option because the existing behavior was not what we want most of the time.

PR #323, opened 2019-05-15
Fix newrelic instrumentation

davisagli:
It needs a bit of help to work with Channels and with rq.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "webpack:serve": "webpack-dev-server --config webpack.dev.js",
-    "django:serve": "daphne metadeploy.asgi:application",
+    "django:serve": "python manage.py runserver",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "webpack:serve": "webpack-dev-server --config webpack.dev.js",
-    "django:serve": "python manage.py runserver",
+    "django:serve": "daphne metadeploy.asgi:application",
```
wlonk: Why change the method of running in local development? This will require us to manually restart the server on code changes, right?

davisagli: @wlonk I was missing that the channels app provides its own runserver command. I'll switch back to that but need to revisit where the newrelic agent gets initialized.

davisagli: @wlonk updated.

PR #324, opened 2019-05-16
[change-preflight-shape] Handle the new shape of Preflight results

wlonk:
They no longer have a list of results for each step ID, just a single results dict.

See https://trello.com/c/3TOulyHK/97-1-do-not-show-preflight-with-warnings-modal-if-warning-step-is-not-checked

giraffe


wlonk:
@duggiemitchell To answer your general process question, we generally approve, at least verbally, with or without coverage, sometimes with things like "this looks good, assuming you get the tests in" or whatever. Of course, a formal GitHub approval will be rendered stale by pushing a new commit with tests, so a formal approval doesn't matter much. And we try to do proper TDD, writing tests first, when we can! But I usually fail to do that on the JS side of things.

duggiemitchell:
> @duggiemitchell To answer your general process question, we generally approve, at least verbally, with or without coverage, sometimes with things like "this looks good, assuming you get the tests in" or whatever. Of course, a formal GitHub approval will be rendered stale by pushing a new commit with tests, so a formal approval doesn't matter much. And we try to do proper TDD, writing tests first, when we can! But I usually fail to do that on the JS side of things.

OK thanks!  That's good to know 👍.

PR #329, opened 2019-05-23
[resolve-warning] Make preflight warnings clearer.

jgerigmeyer:
https://trello.com/c/o4i1KfSQ

german shepherd

### DIFF
--- a/src/js/components/plans/stepsTable/index.js
+++ b/src/js/components/plans/stepsTable/index.js
@@ -209,6 +209,9 @@ class StepsTable extends React.Component<Props, State> {
               <NameDataCell
                 preflight={preflight}
                 job={job}
+                selectedSteps={
+                  hasValidToken && hasReadyPreflight ? selectedSteps : undefined
+                }
                 activeJobStep={activeJobStepId}
                 togglePanel={this.togglePanel}
                 expandedPanels={expandedPanels}
--- a/src/js/components/plans/stepsTable/nameDataCell.js
+++ b/src/js/components/plans/stepsTable/nameDataCell.js
@@ -49,7 +49,7 @@ class NameDataCell extends React.Component<Props> {
     const { id } = item;
     const isActive = Boolean(activeJobStep && id === activeJobStep);
     const result = currentJob && currentJob.results && currentJob.results[id];
-    const isSelected = !selectedSteps || selectedSteps.has(id);
+    const showErrorColors = !selectedSteps || selectedSteps.has(id);
     let hasError = false;
     let hasWarning = false;
     let optionalMsg = '';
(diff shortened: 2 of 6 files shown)


jgerigmeyer:
Notable changes:

- When any step-with-warning is selected, change CTA text to "View Warnings to Continue Installation"
- When step-with-warning is not selected, remove yellow color on left side of step row

One warning selected:

Screenshot 2019-05-23 17 52 54

Two warnings selected:

Screenshot 2019-05-23 17 53 00

No warnings selected:

Screenshot 2019-05-23 17 53 06

src/js/components/plans/stepsTable/index.js
```diff
@@ -209,6 +209,9 @@ class StepsTable extends React.Component<Props, State> {
               <NameDataCell
                 preflight={preflight}
                 job={job}
+                selectedSteps={
+                  hasValidToken && hasReadyPreflight ? selectedSteps : undefined
```
wlonk: Does the difference between `selectedSteps=undefined` and `selectedSteps=[]` matter? If not, can we prefer `[]` as the empty case?

src/js/components/plans/stepsTable/nameDataCell.js
```diff
@@ -48,6 +49,7 @@ class NameDataCell extends React.Component<Props> {
     const { id } = item;
     const isActive = Boolean(activeJobStep && id === activeJobStep);
     const result = currentJob && currentJob.results && currentJob.results[id];
+    const isSelected = !selectedSteps || selectedSteps.has(id);
```
wlonk: This logic surprises me; can you explain? If there are no selected steps, this cell is selected?

src/js/components/plans/stepsTable/index.js
jgerigmeyer: You can only select/deselect steps if you have a valid token and preflight -- otherwise we're not showing the checkbox column at all. In that case, I want to consider every step selected for the purposes of displaying errors/warnings. So I'm making a distinction between "there is no such thing as a selected step yet" and "no steps are selected."

src/js/components/plans/stepsTable/nameDataCell.js
jgerigmeyer: See comment above. Maybe renaming would be helpful?

wlonk: Yeah, probably; `isSelectedIfSelectedPossible` is clunky, though 😆

mirisuzanne:
I think we need a disabled (or warning?) button state if we're asking you  to do something else before you continue.

jgerigmeyer:
> I think we need a disabled (or warning?) button state if we're asking you to do something else before you continue.

@mirisuzanne We're not changing the functionality or asking you to do anything else -- just making the text of the CTA button match what actually happens when you click it: you get the modal with the warnings listed.

jgerigmeyer:
@mirisuzanne or @wlonk Can one of you re-review and approve for merge?

PR #330, opened 2019-05-24
[product-tab-link] Link directly to a product-category tab.

jgerigmeyer:
https://trello.com/c/gktsFzhl

panda


src/js/components/products/list.js
```diff
@@ -63,11 +66,35 @@ class ProductsList extends React.Component<Props, State> {
...
+    if (hasCategories && !activeProductsTab) {
+      // make the first category the hash url
+      urlHash = prettyHashUrl(productCategories[0]);
+    } else {
+      // make saved item  in storage
+      urlHash = prettyHashUrl(activeProductsTab);
+    }
+    // what if we load the page and the hash is different than the
+    const location = {
+      pathname: '/products',
```
jgerigmeyer: I think you can remove `pathname` here -- if you just pass in `hash`, it will only update the hash without changing the full path.

src/js/components/products/list.js
```diff
@@ -63,11 +66,35 @@ class ProductsList extends React.Component<Props, State> {
...
+      hash: urlHash,
+    };
+    this.setState({ hash: urlHash });
+    history.push(location);
```
jgerigmeyer: Use `history.replace()` instead, so it does not add a new history item?

src/js/components/products/list.js
```diff
@@ -63,11 +66,35 @@ class ProductsList extends React.Component<Props, State> {
       } else {
         window.sessionStorage.removeItem('activeProductsTab');
       }
+      this.setState({ hash: prettyHashUrl(category) });
```
jgerigmeyer: I'm not clear on what this is for -- it looks like it's not being used anywhere, just set.

src/js/utils/helpers.js
```diff
@@ -0,0 +1,2 @@
+// @flow
+export const prettyHashUrl = str => str.replace(/\s+/g, '-').toLowerCase();
```
jgerigmeyer: This looks nice, but I'm not sure it's robust enough? We need a way to consistently both encode and decode the category name (string) to a url hash. If we just replace, how can we consistently decode back to the category name?

My initial thought would be to use `encodeURIComponent` and `decodeURIComponent` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent), but that won't result in very pretty urls when the name has spaces in it...

Maybe we first replace all spaces with `-` to cover the basic use-case, then also run it through `encodeURIComponent` to be sure we're encoding any other special characters. It'll just mean that going the other way we'll have to `decodeURIComponent` and then compare the result against the initial category names _with spaces replaced by dashes_.

jgerigmeyer: Turns out I was wrong -- hashes can have any chars except e.g. `%` and `#` -- and even there, it looks like Chrome doesn't mind those.

PR #331, opened 2019-05-24
📦 Update cumulusci

davisagli:


jgerigmeyer:
Hm, we now have a dependency version conflict between boto3 and cumulusci:

> ERROR: botocore 1.12.155 has requirement urllib3<1.25,>=1.20; python_version >= "3.4", but you'll have urllib3 1.25.2 which is incompatible.

> ERROR: cumulusci 2.5.0 has requirement urllib3==1.25.2, but you'll have urllib3 1.24.3 which is incompatible.

davisagli:
@jgerigmeyer I'm confused, because this commit in botocore that should allow urllib3 1.25 claims to be included in the botocore 1.12.155 tag: https://github.com/boto/botocore/commit/3a6bd282307ff91cbd04f2a7dfa6369c08015d2f

davisagli:
@jgerigmeyer possible fix in https://github.com/SFDO-Tooling/MetaDeploy/pull/335

PR #332, opened 2019-05-24
[omit-unlisted] Omit unlisted Products entirely from List view

wlonk:
See https://trello.com/c/y4EbqfHQ/154-omit-islisted-false-from-products-api-unless-requested-by-specific-slug

red wolves


wlonk:
I think this just needs functional testing.

jgerigmeyer:
@wlonk If I'm not mistaken, I think we want to do this for versions and plans as well? We also need minor front-end adjustments to allow fetching an individual product by direct URL.

jgerigmeyer:
@wlonk Functionally, on this branch I'm still seeing all products returned in the list, even with `is_listed: False`

wlonk:
My best guess is that this is something to do with `is_listed` in the serializer combining the field and the "can you only see this because of your org type being in the allowed list", but I'll need to see what you're seeing to confirm that.

jgerigmeyer:
@wlonk Just to document what we discussed this morning:

Todo:

- [ ] make sure `is_listed: False` is filtered out of list GET for products, versions, plans
- [ ] make sure an individual product/version/plan can be fetched even if `is_listed: False` (clarify endpoint for this)

wlonk:
Crucial to my understanding of point two: individual resources are not fetched by ID, but rather by a set of filters on the list endpoint that are expected to narrow it down to 0 or 1 results. Which makes this all a lot fuzzier.

jgerigmeyer:
> Crucial to my understanding of point two: individual resources are not fetched by ID, but rather by a set of filters on the list endpoint that are expected to narrow it down to 0 or 1 results. Which makes this all a lot fuzzier.

That's correct. For products, we have `product.slug`. For versions, we have `product.id` and `version.label`. For plans, we have `product.id`, `version.id`, and `plan.slug`.

jgerigmeyer:
@wlonk Todo list edited to add filtering out items with `is_listed_by_org_only: True`.

wlonk:
@jgerigmeyer if you would like to check that this behaves as you expect, I have pushed fixes that should.

To get a single entry or 404, hit `.../api/{plan,version,product}/get_one/?...` with appropriate filters, per discussion.

mirisuzanne:
[Omit `is_listed: False` from products/versions/plans list APIs.](https://trello.com/c/y4EbqfHQ/154-omit-islisted-false-from-products-versions-plans-list-apis)

jgerigmeyer:
@davisagli I think this is ready for merge, but I'd appreciate your code and/or functional review if you get a chance.

PR #333, opened 2019-05-24
[average-time-note] 🛢 Show disclaimer tooltip with average-install-time for production orgs.

jgerigmeyer:
https://trello.com/c/GjBjirTb

fox


jgerigmeyer:
This adds the tooltip for production orgs, but does not change how we calculate average times (that is, we are not calculating averages based on org-type).

Screenshot 2019-05-24 12 32 25
----
Screenshot 2019-05-24 12 32 33

metadeploy/api/serializers.py
```diff
@@ -95,9 +96,13 @@ class Meta:
             "valid_token_for",
             "org_name",
             "org_type",
+            "is_production_org",
             "is_staff",
         )
 
+    def get_is_production_org(self, obj):
+        return obj.full_org_type == "Production"
```
jgerigmeyer: @wlonk I don't like how these values are hardcoded here (and also in the `full_org_type` property itself). I think it'd be better to move the ORG_TYPES tuple to constants.py (I guess making it an enum or something) and expose it to the front-end for comparison.

metadeploy/api/models.py
```diff
@@ -651,6 +651,7 @@ class Job(HashIdMixin, models.Model):
     org_id = models.CharField(null=True, blank=True, max_length=18)
     org_name = models.CharField(blank=True, max_length=256)
     org_type = models.CharField(blank=True, max_length=256)
+    full_org_type = models.CharField(null=True, blank=True, max_length=256)
```
davisagli: I was going to say we don't need this field here since the Job serializer can just get the value from the user, but I think it'll come in handy to be able to filter jobs based on this, so this is fine.

metadeploy/api/serializers.py
wlonk: Lemme go do that cleanup.

jgerigmeyer:
> I want to summarily approve this because of the cute animal picture, but I do have one caution: given that we have three types of orgs, having a "is it Type A" boolean feels a bit off to me; it works for the specific use-case here, and maybe we only change it if we ever have to check for a different kind, but it does feel, usually, like if you have a "pick from three" type, you should expose that and let the client test against the value.

I think there are two levels of cleanup here. First, we shouldn't be hardcoding the values directly in the models to begin with. Then, if we like, we can expose constants to the front-end for comparison instead of this boolean. I feel more strongly about the former than the latter, but am good with both if you prefer.

PR #338, opened 2019-05-28
[pagination] Pagination

wlonk:
See https://trello.com/c/0LSugzU3/17-3-as-an-admin-i-want-pagination-to-help-organize-the-product-list

sloth bear cub


src/js/components/products/list.js
```diff
@@ -68,6 +69,24 @@ class ProductsList extends React.Component<Props, State> {
...
   };
 
+  isBottom = el => el.getBoundingClientRect().bottom <= window.innerHeight;
+
+  trackScrolling = () => {
+    const wrappedElement = document.getElementById('app');
+
+    if (wrappedElement && this.isBottom(wrappedElement)) {
+      this.setState({ fetchMoreProducts: true });
+    }
```
duggiemitchell: @jgerigmeyer trying to write tests for this functionality and looks like you have to sort of hack this together per: https://stackoverflow.com/questions/38656541/change-element-size-using-jest.

Having difficulty getting this to work, any idea if this is the right approach from a testing perspective?  Do you know if there are examples of tests somewhere that mocks the window size, position, etc for me to look at?

docs/api/plans.rst
```diff
@@ -17,23 +17,28 @@ List
...
+         "old_slugs": [],
+         "steps": [],
+         "is_allowed": true,
+         "is_listed": true,
+         "not_allowed_instructions": null,
+         "requires_preflight": true
+       },
+       ...
+     ]
+   }
```
jgerigmeyer: @wlonk Is there an easy way to _only_ paginate products, not other lists?

src/js/components/products/list.js
jgerigmeyer: @duggiemitchell Hm, it seems a bit like an anti-pattern for this component to listen to events on a DOM element external to itself. I'd much rather use the React `onScroll` synthetic event (https://reactjs.org/docs/handling-events.html), but I think that would require the scrolling element to be within this component... where right now the entire `html` element is scrolling.

@stacyk What do you think about have the tab-contents scroll (currently `div.slds-grid.slds-wrap`), instead of the entire page?

duggiemitchell: @jgerigmeyer makes sense.  are you suggesting we do something like `overflow: scroll` on the above class?

jgerigmeyer: @duggiemitchell I think we should bind the scroll to an element within the component, yes. If that works without changing what scrolls, it's fine by me. But I wonder if the UX will be clearer to have the scroll internal as well? Happy to defer to @mirisuzanne or @stacyk or @SondraE on that question.

docs/api/plans.rst
wlonk: Yes. Is that preferable? I suppose it is! I'll add that in.

src/js/components/products/list.js
stacyk: It's hard to say for sure without testing it ... the installation pages scroll the full page if needed (except horizontally, the table will scroll if pages isn't wide enough). I don't know if have a mix of both would create confusion or people may not notice. A benefit would include having the tab menu and header available at all times.

This doesn't provide an answer other than I'm not sure it will make a big usability difference. @mirisuzanne or @SondraE do you have an opinion?

jgerigmeyer:
@wlonk From the user's perspective, I think ideally we want to paginate-by-category (that is, returning the first 25 from each category). How would you like to approach that?

mirisuzanne:
[As an Admin, I want pagination to help organize the product list.](https://trello.com/c/0LSugzU3/17-as-an-admin-i-want-pagination-to-help-organize-the-product-list)

src/js/components/products/list.js
```diff
@@ -88,20 +90,32 @@ class ProductsList extends React.Component<Props, State> {
...
+    const scrolledToBottom =
+      Math.ceil(scrollTop + clientHeight) >= scrollHeight;
 
-    if (wrappedElement && ProductsList.isBottom(wrappedElement)) {
-      this.setState({ fetchMoreProducts: true });
+    if (scrolledToBottom) {
+      this.setState({ fetchingProducts: true });
+      this.props.dofetchProducts().then(() => {
+        this.setState({ fetchingProducts: false });
+      });
```
duggiemitchell: @jgerigmeyer I think I'm missing something here with setting up this action in the component; this is throwing a TypeError: _`this.props.dofetchProducts is not a function
    at HTMLDocument.`

jgerigmeyer: @duggiemitchell There are three different spellings of (I think) the same function in this file:

- `doFetchProducts`
- `dofetchProducts`
- `doFetchMoreProducts`

src/js/components/products/list.js
```diff
@@ -145,6 +179,22 @@ class ProductsList extends React.Component<Props, State> {
...
+                <div
+                  role="status"
+                  className="slds-spinner slds-spinner_small slds-spinner_brand spinner-container"
+                >
+                  <span className="slds-assistive-text">Loading</span>
+                  <div className="slds-spinner__dot-a" />
+                  <div className="slds-spinner__dot-b" />
+                </div>
+                <span>{i18n.t('Loading…')}</span>
+              </div>
```
jgerigmeyer: Can we use the `<Spinner />` component that we're using elsewhere?

https://react.lightningdesignsystem.com/components/spinners/

src/js/components/products/list.js
```diff
@@ -157,8 +207,12 @@ const select = (appState: AppState) => ({
   productsByCategory: selectProductsByCategory(appState),
 });
 
-const WrappedProductsList: React.ComponentType<InitialProps> = connect(select)(
-  ProductsList,
-);
+const actions = () => ({
+   doFetchProducts: fetchProducts
+});
```
jgerigmeyer: This can just be an object, not a function:

```js
const actions = {
  doFetchProducts: fetchProducts,
};
```

src/js/store/products/selectors.js
```diff
@@ -59,6 +59,17 @@ const selectProductCategories: AppState => Array<string> = createSelector(
...
+  appState.products.categories;
+
+const selectProductCount: AppState => Array<string> = createSelector(
+  productCount,
+  count => {
+    const categoryIdCount = new Set();
+    categoryIdCount.add(count);
+    return categoryIdCount;
+  },
+);
```
duggiemitchell: @jgerigmeyer if you run this, you will see that the `count` is not being assigned to the `productCount` prop.

src/js/store/products/actions.js
```diff
@@ -3,13 +3,13 @@
...
 import apiFetch, { addUrlParams } from 'utils/api';
-import type { Product, Version } from 'store/products/reducer';
+import type { Category, Product, Version } from 'store/products/reducer';
 import type { Plan } from 'store/plans/reducer';
 
 type FetchProductsStarted = { type: 'FETCH_PRODUCTS_STARTED' };
 type FetchProductsSucceeded = {
   type: 'FETCH_PRODUCTS_SUCCEEDED',
-  payload: Array<Product>,
+  payload: [Array<Product>, Array<Category>],
```
jgerigmeyer: Any reason to use an array of arrays here, instead of an object with named keys, e.g.:

```js
{
  products: Array<Product>,
  categories: Array<Category>,
}
```

src/js/store/products/actions.js
```diff
@@ -87,9 +91,39 @@ export type ProductsAction =
   | FetchPlanSucceeded
   | FetchPlanFailed;
 
+export const fetchMoreProducts = (
+  category: string,
+  page: number,
+): ThunkAction => dispatch => {
+  dispatch({ type: 'FETCH_MORE_PRODUCTS_STARTED' });
+  const baseUrl = `${window.api_urls.product_list()}?category=${category}&page=${page}`;
```
jgerigmeyer: Use `addUrlParams` helper.

src/js/store/products/actions.js
```diff
@@ -98,7 +132,19 @@ export const fetchProducts = (): ThunkAction => dispatch => {
         error.response = response;
         throw error;
       }
-      return dispatch({ type: 'FETCH_PRODUCTS_SUCCEEDED', payload: response });
+      let categories = [];
+      let products = [];
+      let count = [];
+      response.forEach(({ first_page, id, title }) => {
+        products = products.concat(first_page.results);
+        categories = categories.concat({ first_page, id, title });
```
jgerigmeyer: I don't think `first_page` is needed here? That is, I'm not sure why we need to store it along with the category.

src/js/store/products/actions.js
```diff
@@ -98,7 +132,19 @@ export const fetchProducts = (): ThunkAction => dispatch => {
...
+        count = count.concat({ [id]: first_page.count });
```
jgerigmeyer: Why do we need to store the `count` at all? I don't see it being used anywhere...

It's the total # of products in a given category, so it can't really be used to determine the page-number. Instead I'd suggest we store the value of `next`, and use that to fetch more products. If it's `null`, we know there aren't any more to fetch.

src/js/store/products/reducer.js
```diff
@@ -41,15 +41,33 @@ export type Product = {
...
   action: ProductsAction,
 ): ProductsState => {
   switch (action.type) {
     case 'FETCH_PRODUCTS_SUCCEEDED':
-      return { ...products, products: action.payload };
+      return {
+        ...products,
+        products,
+        categories,
+        count,
```
jgerigmeyer: These new values aren't defined anywhere. Based on the payload you defined in the `actions.js` file, `action.payload` is `[products, categories, count]`. In which case this would need to be:

```js
      return {
        ...products,
        products: action.payload[0],
        categories: action.payload[1],
        count: action.payload[2],
      };
```

But as I said in a comment above, I think that structure is overly opaque (and I don't think storing `count` is helpful).

src/js/store/products/selectors.js
```diff
@@ -43,11 +46,29 @@ const selectProductsByCategory: AppState => ProductsMapType = createSelector(
...
+const productCount = (appState: AppState): ProductsState =>
+  appState.products.categories;
+
+const selectProductCount: AppState => Array<string> = createSelector(
+  productCount,
+  count => {
+    const categoryIdCount = new Set();
+    categoryIdCount.add(count);
+    return categoryIdCount;
+  },
```
jgerigmeyer: Without having a clearer sense of how the State is structured and how the components need to access that state, I'm not sure what these selectors need to do/return.

src/js/components/products/list.js
```diff
@@ -83,6 +104,49 @@ class ProductsList extends React.Component<Props, State> {
...
+    });
+  };
+
+  componentDidMount() {
+    document.addEventListener('scroll', this.handleOnScroll);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('scroll', this.handleOnScroll);
+  }
```
jgerigmeyer: I still don't love having a component bind to an event external to the component, but maybe this is okay.

metadeploy/api/management/commands/populate_data.py
```diff
@@ -275,7 +275,7 @@ def create_preflight_expiry_job(self):
         )
 
     def create_eda(self):
-        sf_category = ProductCategory.objects.get(title="salesforce.org products")
+        sf_category = ProductCategory.objects.get(title="Salesforce.org Products")
```
jgerigmeyer: SLDS is moving away from all-caps text, and this PR upgrades design-system-react to a version where they've removed all-caps from the tab labels:

Screenshot 2019-06-14 16 01 35

We could force them uppercase in display, or add a `display` field to the model, or just display them as they're defined in the database (possibly with a data migration?).

docs/api/products.rst
```diff
@@ -16,84 +16,89 @@ List
...
-       "is_allowed": true,
-       "is_listed": true,
-       "order_key": 0,
-       "not_allowed_instructions": null
-     },
-     ...
-   ]
+       ...
+     ]
+   }
```
jgerigmeyer: @wlonk This looks good, but docs are missing for the new `/categories/` endpoint.

src/js/components/products/list.js
```diff
@@ -145,6 +203,14 @@ class ProductsList extends React.Component<Props, State> {
               />
             ) : null}
             {contents}
+            {fetchingProducts ? (
+              <div className="slds-align_absolute-center">
+                <span className="slds-is-relative slds-m-right_large">
+                  <Spinner variant="brand" size="small" />
+                </span>
+                {i18n.t('Loading…')}
+              </div>
```
jgerigmeyer: @stacyk This spinner needs a bit of style cleanup:

Screenshot 2019-06-14 16 06 07

wlonk:
@jgerigmeyer that's a big eternal problem; I don't have a good answer to it. We should talk, eh?

test/js/store/products/actions.test.js
```diff
@@ -63,6 +73,76 @@ describe('fetchProducts', () => {
   });
 });
 
+describe('fetchMoreProducts', () => {
+  describe('success', () => {
+    test('GETs next products list for category', () => {
+      const store = storeWithApi({});
+      const url = `${window.api_urls.product_list()}?category=30&page=2`;
```
jgerigmeyer: Probably use `addUrlParams` helper here and throughout. See fetchProduct test below: https://github.com/SFDO-Tooling/MetaDeploy/pull/338/files#diff-9790db778d0c454c9dd68cfff28aba7eR158

test/js/store/products/actions.test.js
```diff
@@ -63,6 +73,76 @@ describe('fetchProducts', () => {
...
+      fetchMock.getOnce(url, mockResponse);
+      const started = {
+        type: 'FETCH_MORE_PRODUCTS_STARTED',
+        payload: { url, id },
+      };
+      const succeeded = {
+        type: 'FETCH_MORE_PRODUCTS_SUCCEEDED',
+        payload: { products: nextProducts, category: id, next: null },
+      };
+      expect.assertions(1);
```
jgerigmeyer: My convention (obviously just personal preference) is to have an empty line between setup/execution lines and assertions -- so always a new line before and `expect` statement.

test/js/store/products/actions.test.js
```diff
@@ -63,6 +73,76 @@ describe('fetchProducts', () => {
...
+      return store.dispatch(actions.fetchMoreProducts({ url, id })).then(() => {
+        expect(store.getActions()).toEqual([started, succeeded]);
+      });
+    });
+  });
+  describe('error', () => {
```
jgerigmeyer: Missing newline (between each test or suite).

metadeploy/api/serializers.py
```diff
@@ -213,6 +217,42 @@ class Meta:
...
+        """
+        We expect this to always be None, because we know we're returning the first
+        page.
+        """
+        return None
+
+    def get_first_page(self, obj):
+        paginator = ProductPaginator()
+        page = paginator.paginate_queryset(
+            obj.product_set.all(), self.context["request"]
```
jgerigmeyer: @wlonk This appears to bypass the newly-added `FilterAllowedByOrgMixin` used by the `ProductViewSet`, introducing a regression where `is_listed: False` products are again included in data returned by the API.

metadeploy/api/management/commands/populate_data.py
davisagli: Just leave it up to individual deployments.

test/js/components/products/list.test.js
```diff
@@ -221,4 +227,42 @@ describe('<Products />', () => {
...
+        notFound: [],
+        categories: [{ id: 1, title: 'salesforce', next: null }],
+      },
+    };
+    const newProducts = {};
+    const { baseElement } = setup(initialState);
+    console.log(window.pageYOffset);
+    fireEvent.scroll(baseElement, { y: 1000 });
+    console.log(window.pageYOffset);
+  });
```
duggiemitchell: @jgerigmeyer I'm sort of stumped on what the setup should be to simulate the element scrolling.

jgerigmeyer: @duggiemitchell Oof, you're not wrong. This is a difficult interaction to test. Rather than try to mimic a scroll event (which essentially tests that the third-party `withScroll` higher order fn is doing what we expect), I think it'd be easier to mock that out so we can manually pass the `y` value we want. I've pushed an example passing test that does this (https://github.com/SFDO-Tooling/MetaDeploy/commit/319998c18005cd1d8cac43b8cb364098d715c0fb) -- take a look and let me know if you have any questions.

duggiemitchell: Thanks! this helps.  I'm curious why the coverage report leaves out some directories (such as this test) and if that has an effect on the coverage summary.
image

But you can still find it by the url
image

jgerigmeyer: @duggiemitchell Yeah, each time it only produces coverage for the tests that are actually run, and by default `test:watch` only runs test for files that have local changes (since the last commit). You can also filter the test run using custom file/test name patterns, or only run failed tests, etc. But it doesn't necessarily wipe out the entire `coverage/` directory every time, which is why you can find (outdated) report files by URL/path. But the updated files from the last run will always be linked from `coverage/index.html`, so that's a good place to start.

I sometimes find that a `test:watch` run doesn't actually output coverage results (I think related to: https://github.com/facebook/jest/issues/7331).

The safest is to do a clean `yarn test` run and then open `coverage/index.html` -- that will get you the full updated report.

test/js/components/products/list.test.js
```diff
@@ -236,6 +236,17 @@ describe('<Products />', () => {
...
+      test('changes active tab onSelect', () => {
+        window.sessionStorage.setItem('activeProductsTab', 'salesforce');
+        const { getByText } = setup(initialState);
+        const activeTab = getByText('salesforce');
+        const inactiveTab = getByText('community');
+
+        expect(activeTab).toHaveClass('slds-active');
+        fireEvent.click(inactiveTab);
+        expect(inactiveTab).toHaveClass('slds-active');
+      });
```
duggiemitchell: @jgerigmeyer I cleaned up some of the previous tests.  This is what's most confusing, figuring out how your assertions can pass, but it's not actually executing the code as expected such as this test for changing state in `handleSelect`

https://github.com/SFDO-Tooling/MetaDeploy/blob/pagination/src/js/components/products/list.js#L87

jgerigmeyer: Yeah, that is a bit weird. I think it _is_ executing the right code, but probably getting confused because of the unmocked `history.replace()` call (which the test runner might not know what to do with). If I switch the order of these lines, it seems to work as expected. 🤷‍♂

wlonk:
@jgerigmeyer and @duggiemitchell, my punch list items are in. I can review when ready.

src/js/components/products/list.js
```diff
@@ -119,6 +119,7 @@ class ProductsList extends React.Component<Props, State> {
     if (y === prevProps.y || fetchingProducts) {
       return;
     }
+    /* istanbul ignore next */
```
duggiemitchell: @jgerigmeyer I think I may have asked this already, spent a lot of time trying to figure out how to test for this line, but curious about the criteria which lines to ignore for testing.

src/js/components/products/list.js
```diff
@@ -129,6 +130,7 @@ class ProductsList extends React.Component<Props, State> {
     // Fetch more products if within 100px of bottom of page...
     const scrolledToBottom = scrollHeight - Math.ceil(y + clientHeight) <= 100;
 
+    /* istanbul ignore else */
     if (scrolledToBottom) {
       this.maybeFetchMoreProducts();
```
duggiemitchell: @jgerigmeyer ditto on this (ignored line).

src/js/components/products/list.js
jgerigmeyer: Great question. It’s always a bit of a judgment call, but I do have a few general rules I try to follow. One is that I’m quick to ignore a line/branch if the branch itself is impossible or very very unlikely to be reproduced in practice. That is, when the code is written (perhaps overly) defensively, I’m more likely to ignore coverage rather than try to bend over backwards to get coverage. That’s sort of how I felt this line, where there’s a simple fallback in case `document.documentElement` is missing. I don’t know how difficult it would be to get coverage on that line (it can be a pain to mock or work with DOM globals in a node environment like jest/jsdom), but it didn’t seem like a valuable thing that really needed to be tested.

src/js/components/products/list.js
jgerigmeyer: Yeah... this is a bit less clear. I guess I’m more likely to ignore something if it’s a straightforward/obvious piece of code. It wouldn't be too difficult to test this… but at the same time, does it really need to be tested? It’s pretty clear what will happen in this case, and I’m less likely to test the negative. (But probably I should have just tested it; I was getting lazy.)

PR #350, opened 2019-06-06
📦 🛢 Switch to new format for configuring preflight checks

davisagli:
image

This refactors how preflight checks are defined and evaluated including:
- Replace `preflight_flow_name` with a `preflight_checks` JSONField to the Plan model (for configuring plan-level checks) and a `checks` list in the task_config for a Step model.
- Move the instantiation of a FlowCoordinator to a `run` method of the Job & PreflightResult models, because the isinstance checks were a code smell that it should be dispatched as a method
- Use the new `PreflightFlowCoordinator` in CumulusCI to actually perform the checks
- Adjusted `PreflightFlowCallback` since `PreflightFlowCoordinator` now returns results in a format closer to what the MetaDeploy frontend expects.
- Update the sample data added by the populate_data command to the new format.

There's no automatic migration of preflight_flow_name. That's fine, because we aren't using it yet.


metadeploy/adminapi/api.py
```diff
@@ -73,9 +73,16 @@ class PlanTemplateViewSet(AdminAPIViewSet):
     model_name = "PlanTemplate"
 
 
+class PlanFilter(filters.FilterSet):
+    class Meta:
+        model = models.Plan
+        exclude = ("preflight_checks",)
```
wlonk: I generally prefer to explicitly include fields on things like filters and serializers; I'm not gonna say this is a blocking concern, but it is a concern.

requirements/base.txt
```diff
@@ -1,10 +1,10 @@
 -e git+https://github.com/SFDO-Tooling/sfdo-template-helpers.git@v0.7.0#egg=sfdo-template-helpers
+-e git+https://github.com/SFDO-Tooling/CumulusCI.git@feature/preflight#egg=cumulusci
```
wlonk: And we'll remember to replace this with a release version, yeah?

metadeploy/adminapi/api.py
davisagli: Yeah, I should probably revisit this entire module with that in mind. It _is_ restricted to only admins on our network, which makes me less concerned. The immediate blocker I was trying to get past is that django-filters didn't know how to handle the JSONField.

requirements/base.txt
davisagli: Yeah, this is just until the change gets merged and released in CumulusCI

davisagli:
@jgerigmeyer I don't think it's actually a problem in this case, but just because I submit something for final review doesn't mean I'm necessarily ready to merge it yet. In this case I was planning to do one more check to make sure there aren't any existing plans using preflight flows on production.

jgerigmeyer:
@davisagli Thanks for clarifying -- my bad! I assume there's only an auto-deploy to devstaging, but I'll let you merge from here on out.

davisagli:
@jgerigmeyer yeah that's true, it doesn't auto-deploy to production, so I still would have had a chance

PR #352, opened 2019-06-07
Optimize expiration of tokens

davisagli:
The previous approach was doing a separate query to delete the SocialToken for each user (except users with recent logins or running jobs) and was taking longer and longer as we accumulate users. This changes it so we're primarily querying tokens (which should always be a fairly short list) and then doing separate queries for each token to confirm the user doesn't have running jobs.

I imagine it's possible to do it as a single query but this is probably good enough.

### DIFF
--- a/metadeploy/api/jobs.py
+++ b/metadeploy/api/jobs.py
@@ -23,6 +23,7 @@ from glob import glob
 from itertools import chain
 
 import github3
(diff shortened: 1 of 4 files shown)


davisagli:
The New Relic graph from when this was deployed is fun.

Screen Shot 2019-06-10 at 10 06 11 AM

PR #356, opened 2019-06-10
[api-errors] Display API errors.

jgerigmeyer:
anteater


jgerigmeyer:
@davisagli @wlonk I need to fill in a few tests, but I'd love to get feedback on this PR (code and/or functionality). Right now, it's displaying errors for any response 4xx or 5xx except 404. It looks for `response.detail` or `response.non_field_errors`, and then falls back to the `statusText`. The message looks like the following (the top line is the same for any error, while the second line is the info from the server):

Screenshot 2019-06-10 17 34 36

The error messages can be manually dismissed, and in keeping with the SLDS docs they disappear on navigation:

> Do not keep toasts after the user navigates away. A toast displays status that’s only relevant to the current page they are on; if they move to a different page, the toast message is no longer relevant.

jgerigmeyer:
[As an Admin, I want to see API errors (e.g. "No valid preflight").](https://trello.com/c/DmfdVJCB/85-as-an-admin-i-want-to-see-api-errors-eg-no-valid-preflight)

davisagli:
This works for me.

src/js/components/apiErrors.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+
+import * as React from 'react';
+import Toast from '@salesforce/design-system-react/components/toast';
+import ToastContainer from '@salesforce/design-system-react/components/toast/container';
```
davisagli: I'm a little disappointed that this isn't called the Toaster.

src/js/components/apiErrors.js
```diff
@@ -0,0 +1,49 @@
...
+const ErrorToast = ({
+  error,
+  doRemoveError,
+}: {
+  error: ErrorType,
+  doRemoveError: removeError,
+}) => (
+  <Toast
+    labels={{
+      heading: i18n.t("Uh oh. We've encountered an error. You might try "),
```
stacyk: How about "[Please] reload the page" instead of "You might try reloading the page"? It's up to you, but "you might try reloading the page" makes it sound like we don't really know what to tell you.

jgerigmeyer: Well... yeah, we don't really. That is, reloading the page _might_ help, but this is shown for a number of different types of errors -- we can't say for sure. Usually the details displayed below this sentence should be more helpful?

src/js/components/apiErrors.js
```diff
@@ -0,0 +1,49 @@
...
+  doRemoveError,
+}: {
+  errors: Array<ErrorType>,
+  doRemoveError: removeError,
+}) => (
+  <ToastContainer className="half-container">
+    {errors &&
+      errors.map(err => (
+        <ErrorToast key={err.id} error={err} doRemoveError={doRemoveError} />
+      ))}
```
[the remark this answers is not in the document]
jgerigmeyer: Good point. I added that later so I wouldn't have to add the `errors` array to all the tests... 🙄

jgerigmeyer:
> Why have `getState` and `{ apiFetch }` been taken out of the various fetch functions?

We just weren't using `apiFetch` in any way that required it being passed in as a magic argument to the actions -- so I decided this was clearer/simpler. 🤷‍♂

wlonk:
@jgerigmeyer yeah, I think this is cleaner, too!

src/js/components/apiErrors.js
jgerigmeyer: @stacyk Updated to: "Uh oh, we've encountered an error. You may need to reload the page." Better?

PR #359, opened 2019-06-11
[circleci-caching] Better CI caching

jgerigmeyer:
pika


jgerigmeyer:
@davisagli 

Before:

Screenshot 2019-06-11 12 08 41

After:

Screenshot 2019-06-11 12 16 53

jgerigmeyer:
@davisagli Also, if you poke around you'll see that the actual cache size is much smaller (70 MB for npm, 77 MB for pyenv), now that we're no longer caching the yarn cache itself.

PR #360, opened 2019-06-11
Don't pay attention to list_for_allowed_by_orgs for plans

davisagli:
This was preventing plans from being listed on the product detail page if the user's org has access via the allowed list org type

(For example, log into https://install.salesforce.org/products/sal with a scratch org)

### DIFF
--- a/metadeploy/api/serializers.py
+++ b/metadeploy/api/serializers.py
@@ -149,7 +149,6 @@ class PlanSerializer(CircumspectSerializerMixin, serializers.ModelSerializer):
     preflight_message = serializers.SerializerMethodField()
     not_allowed_instructions = serializers.SerializerMethodField()
     requires_preflight = serializers.SerializerMethodField()
-    is_listed = serializers.SerializerMethodField()
 
     class Meta:
         model = Plan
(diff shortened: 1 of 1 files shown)


jgerigmeyer:
@davisagli I thought this functionality was explicitly the intention of https://trello.com/c/9D50UD3p -- can you clarify what has changed? That is, why is this not solved using the AllowedList "List for allowed by orgs" boolean?

davisagli:
@jgerigmeyer in the example I linked above, both the product and the plan are using the same AllowedList. The AllowedList allows all scratch orgs and has "List for allowed by orgs" unchecked to make sure that the product does not show up on the homepage just because the user is logged in with a scratch org. However, this also causes the plan to not be listed when an allowed user tries to use it:

Screen Shot 2019-06-11 at 2 07 10 PM

jgerigmeyer:
@davisagli Okay, so with this change, if a plan is allowed only because of org type it will always be listed. 👍

davisagli:
Yes, and that seems fine to me. It was really the product listings that we were concerned about.

PR #361, opened 2019-06-12
Option for preflight checks to hide steps

davisagli:
Possible approach for hiding steps based on preflight checks. I'm looking for some sanity checks before I work on the tests.


metadeploy/api/serializers.py
```diff
@@ -431,6 +430,14 @@ def validate(self, data):
         if invalid_steps:
             raise serializers.ValidationError(_("Invalid steps for plan."))
 
+        if self.instance and "results" in data:
+            # results are read-only except during creation
+            del data["results"]
```
wlonk: I feel like this is a case where I'd `data.pop("results")` and just discard the value. `del` always makes me feel itchy for some reason.

metadeploy/api/serializers.py
```diff
@@ -431,6 +430,14 @@ def validate(self, data):
...
+        elif "results" in data:
+            # make sure results can't be set initially except to hide steps
+            if any(result["status"] != HIDE for result in data["results"].values()):
+                raise serializers.ValidationError(_("Invalid initial results."))
```
wlonk: This whole conditional feels like it wants to be pulled into a well-named helper method, so it's immediately clear what the intent of this chunk of code is, either instead of or in addition to the explanatory comments.

wlonk:
@davisagli I like it generally. Just some small legibility comments.

metadeploy/api/serializers.py
davisagli: Why? `del` does the same thing if you're not using the popped value, no?

wlonk: It does exactly the same thing, I just get itchy when I see `del`; as I said, this is a case where _I'd_ do it that way. It's purely an aesthetic choice, not one I'd impose on you here.

PR #367, opened 2019-06-24
[hotfix-plan-modals] Close plan modals on Confirm.

jgerigmeyer:
ape

I'm not sure when this regression was introduced, but this PR fixes a bug where after confirming the preflight-with-warnings modal, it stays open and overlaps with the click-through-agreement modal (see https://metadeploy-devstaging.herokuapp.com/products/product-with-useful-data/0.3.1/preflight-with-warnings):

Screenshot 2019-06-24 17 39 59

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "eslint-plugin-flowtype-errors": "^4.1.0",
     "eslint-plugin-import": "^2.17.3",
     "eslint-plugin-prettier": "^3.1.0",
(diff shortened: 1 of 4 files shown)


davisagli:
Thanks for the fix. Should we have a test for this?

jgerigmeyer:
> Thanks for the fix. Should we have a test for this?

That... is a very good idea. 👍

PR #386, opened 2019-07-12
[fix-sentry] Fix sentry (and various cleanups).

jgerigmeyer:
- Upgrade to new Sentry SDKs
- Fix CircleCI/greenkeeper bug

hamster or gerbil


requirements/base.txt
```diff
@@ -26,10 +26,7 @@ newrelic==4.20.1.121
 Pillow==6.0.0
 psycopg2-binary==2.8.3
 pyopenssl==19.0.0
-raven==6.10.0
-redis==3.2.1
-rq-scheduler==0.9
-rq==1.0
```
wlonk: Why are we removing `redis`, `rq-scheduler`, and `rq`?

jgerigmeyer: Those are installed as dependencies of other packages -- the only reason we added them here was to pin their versions back.

wlonk: So they're all coming from `django-rq-scheduler` now?

jgerigmeyer: and `django-redis` and `django-rq`, yes

davisagli: We should be pinning all dependencies, not just direct ones. Otherwise the version of the transitive dependencies that you have is dependent on when you ran the installation. pip does not update packages that are already installed unless there's a requirement requiring it to be updated.

jgerigmeyer: Hrm, okay. That seems unfortunate, since there's no good way to enforce that rule. Does pip not have a lockfile option? I guess maybe that's an argument for pipenv?

wlonk: @jgerigmeyer it is very much part of why the community generally is pushing towards poetry or pipenv, yeah.

jgerigmeyer: @wlonk @davisagli In the absence of a lockfile, I guess my main question is how do you decide which dependencies to pin vs which not to. Only pin ones that are directly imported? Of these, I think that's just `rq`?

davisagli: On most of our projects I have all dependencies pinned. That also makes sure that pyup knows to try updating them. I've just used "pip freeze" to get a list and then manually determined which requirements file they should go in. `pipdeptree` can be useful for understanding the dependency structure.

wlonk: There are different philosophies, but I think that the "pin literally everything" one isn't bad (and is really the most justifiable), except that it makes it hard to see what you are requiring directly, and what transitively.

jgerigmeyer: With that approach, how do you know when a transitive dependency is no longer required by a direct dependency?

wlonk: Yeah. You don't. Unless you keep track of intentional dependencies separately and occasionally rebuild the pinned deps. It's bad.

davisagli: The snarky answer is that unlike node, you have few enough dependencies to know what they do and take care of this manually. But Kit's right, it's bad.

jgerigmeyer: 😆 "Fewer than node" -- true! Still... over 100 python dependencies. 😬

PR #403, opened 2019-07-25
📦 Swap logos; change text for button to view plan

davisagli:
This implements a couple requests from our docs team:

1. Swap the positions of the product and company logos (to emphasize salesforce.org branding)
2. Change the button to view a plan to say "View Details" instead of "View Plan"

I'd appreciate review from Oddbird but please hold off on merging since I'd like to use this to walk my new colleague @Br4nd0R through the deployment process.


requirements/test.txt
```diff
@@ -41,7 +41,7 @@ pluggy==0.12.0
 py==1.8.0
 pycodestyle==2.5.0
 pyflakes==2.1.1
-pyparsing==2.4.1
+pyparsing==2.4.0
```
davisagli: 2.4.1 was deleted from PyPI

jgerigmeyer:
@davisagli Do you think a similar change should be applied to MetaShare (and possibly also SFDO-template)?

davisagli:
@jgerigmeyer I think it's fine to keep the status quo for MetaShare and sfdo-template. Most of our tools aren't as public-facing as MetaDeploy.

PR #404, opened 2019-07-25
The non-build jobs belong in the short queue

davisagli:
This tells the rqscheduler to put repeatable jobs in the "short" queue instead of the default queue.

The scheduler seems to ignore the queue name that is configured on the RepeatableJob records. And the way we were passing queue names to rqscheduler before was also ignored.

I confirmed that jobs enqueued using the delay method (such as the actual preflight and flow runs) still end up in the default queue.

This does mean that it's not possible to set up a RepeatableJob that will use the default queue, which is annoying, but not actually a problem for this project.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,6 +1,6 @@
 web: daphne --bind 0.0.0.0 --port $PORT metadeploy.asgi:application
 devworker: python manage.py rqworker default short
-scheduler: python manage.py rqscheduler default short
+scheduler: python manage.py rqscheduler --queue short
 worker: python manage.py rqworker default
 worker-short: python manage.py rqworker short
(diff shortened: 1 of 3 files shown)


davisagli:
Sigh, django-rq-scheduler is not honoring this though.

jgerigmeyer:
Is this something we want on MetaShare and/or project-template as well?

davisagli:
We definitely need short-running maintenance jobs to be in a different queue from long-running builds, whether that's done like this (which relies on the fact that all of MetaDeploy's repeatable jobs are in that category) or by fixing django-rq-scheduler to actually honor the queue named on the RepeatableJob model.

PR #416, opened 2019-08-05
📦 🛢 🐈 Upgrade deps and add missing migration.

jgerigmeyer:
cat


metadeploy/api/migrations/0054_make_plan_template_required.py
```diff
@@ -28,7 +27,7 @@ def forwards(apps, schema_editor):
         None,
         shared_model=PlanTemplate,
         translations_model=PlanTemplateTranslation,
-        related_name=compat.get_remote_field(PlanTemplateTranslation).related_name,
+        related_name=PlanTemplateTranslation.master.field.remote_field.related_name,
```
wlonk: What's motivating this change? A change in the underlying library?

jgerigmeyer: Yes. It broke when #406 was merged with a failing build. See https://github.com/django-parler/django-parler/commit/4166a6022e08a85393ac86708781973abf0e6890

PR #444, opened 2019-09-06
Locate steps in the db by step_num instead of path

davisagli:
path is not guaranteed to be unique because there can be multiple invocations of the same task name in one flow.

I'm keeping the lookup by path when processing preflight results because that's what CumulusCI gives us -- we'll want to fix this too but not as quickly since it requires coordination between packages.


davisagli:
Oops, I obviously have some work left -- will return to this on Monday.

PR #461, opened 2019-10-15
Feature/allow preflight checks on unlisted plans

Br4nd0R:
* `PlanViewSet.preflight_[get|post]()` can now query from all plans instead of excluding those that are unlisted.

# Changes
* Plans that are unlisted can now have preflight checks pass.

### DIFF
--- a/metadeploy/api/tests/views.py
+++ b/metadeploy/api/tests/views.py
@@ -366,6 +366,26 @@ class TestPreflight:
 
         assert response.status_code == 403
 
+    def test_preflight_where_plan_not_listed(
+        self, client, plan_factory, preflight_result_factory
+    ):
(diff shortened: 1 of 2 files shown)


metadeploy/api/tests/views.py
```diff
@@ -366,6 +366,26 @@ def test_post__unallowed(self, client, plan_factory, allowed_list_factory):
 
         assert response.status_code == 403
 
+    def test_preflight_where_plan_not_listed(
+        self, client, plan_factory, preflight_result_factory
+    ):
+        plan = plan_factory()
+        plan.is_listed = False
+        plan.save()
```
davisagli: You can just do `plan = plan_factory(is_listed=False)` I think.

PR #587, opened 2019-11-11
Dependabot 🐈 dependency updates 11-11-2019

Br4nd0R:


davisagli:
@Br4nd0R looks like there are some js lint errors...I'm not sure what the cause is yet but presumably a side effect of one of the js package updates.

Br4nd0R:
@davisagli - I bet I removed/added an unexpected newline when resolving conflicts. Let me see if I can find the issue.

PR #614, opened 2019-11-18
🐈 js / python dependency updates

Br4nd0R:


package.json
```diff
@@ -78,6 +78,8 @@
     "@babel/plugin-proposal-object-rest-spread": "^7.6.2",
     "@babel/preset-env": "^7.7.1",
     "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-react": "^7.6.3",
+    "@testing-library/jest-dom": "^4.2.4",
```
davisagli: These 2 are in here twice and I think that's related to why the build failed.

PR #710, opened 2019-12-09
📦 Weekly Dependency Updates

Br4nd0R:


requirements/base.txt
```diff
@@ -72,9 +72,9 @@ idna==2.8
 incremental==17.5.0
 jmespath==0.9.4
 jwcrypto==0.6.0
-keyring==19.2.0
+keyring==19.3.0
```
davisagli: @Br4nd0R I think keyring can be updated to 20.0.0

PR #902, opened 2020-01-27
Week 5 dependency updates

Br4nd0R:


davisagli:
There are a couple failures we'll need to look into.

davisagli:
@Br4nd0R I think I've fixed the errors here. Could you add the updates for week 6 as well?

PR #976, opened 2020-02-18
Various fixes for MetaDeploy docker

davisagli:
@jofsky I needed these fixes to get things working -- let me know if you want me to explain any of it in more detail.

- Expose the webpack proxy server on port 8080 rather than Django on port 8000.
- Replace node-sass with dart-sass.
- Run the correct commands in start_server.sh
- Restore docs for running populate_social_apps when not using docker


webpack.dev.js
```diff
@@ -24,6 +24,7 @@ module.exports = merge(common, {
   devtool: 'cheap-module-inline-source-map',
   devServer: {
     index: '',
+    host: '0.0.0.0',
     proxy: {
       '**': 'http://localhost:8000',
```
jofsky: since you changed docker-compose.yml's ports do you need to change this as well?

PR #1066, opened 2020-03-12
API and model for loading translations

davisagli:
This adds an API (a PATCH to `/admin/rest/translations/{lang}`) that can be used to update translations for objects stored in the database. The API will be called by a task in CumulusCI (https://github.com/SFDO-Tooling/CumulusCI/pull/1577) that sends translations received from the localization team in this format:

    {
        "context": {
            "slug": {
                "message": "translated text",
                "description": "explanation of how the text is used"
            }
        }
    }

- `context` identifies what object the translation is for (might be something like `npsp:plan:existing_org`)
- `slug` identifies what field on the object the translation is for

The API endpoint stores the translations in a new model called Translation. It then runs `update_all_translations` which goes though all translatable objects to make sure that their django-parler translation models are up to date.

Saving a new Plan will also trigger an update of its translations.


davisagli:
Closing in favor of #1121

PR #1121, opened 2020-03-24
Finish localization

davisagli:
- Update translations with fixes from l10n team
- Add a Translation model to store translations in a separate context from the models that use them, and an API for updating the translations.
- Include product slug in the admin API's product serializer so that CumulusCI can look it up.
- Make sure that newly added Plans and Steps get their translations updated from the Translation model.
- Make sure the SiteProfile is translated when it's serialized.
- A few fixes in frontend code to make sure we're translating complete strings

More details about the API for updating translations:

This adds an API (a PATCH to /admin/rest/translations/{lang}) that can be used to update translations for objects stored in the database. The API will be called by a task in CumulusCI (SFDO-Tooling/CumulusCI#1577) that sends translations received from the localization team in this format:

```
{
    "context": {
        "slug": {
            "message": "translated text",
            "description": "explanation of how the text is used"
        }
    }
}
```

- context identifies what object the translation is for (might be something like npsp:plan:existing_org)
- slug identifies what field on the object the translation is for
- The API endpoint stores the translations in a new model called Translation. It then runs update_all_translations which goes though all translatable objects to make sure that their django-parler translation models are up to date.

Saving a new Plan will also trigger an update of its translations.


metadeploy/adminapi/api.py
```diff
@@ -119,3 +123,36 @@ class AllowedListOrgSerializer(AdminAPISerializer):
...
+class TranslationViewSet(viewsets.ViewSet):
+    """Facilitates adding/updating translated text.
+
+    PATCH /admin/rest/translations/es
+
+    {
+        "context": {
+            "slug": {
+                "message": "En español",
+                "description": "Spanish translation for slug in this context"
```
davidmreed: Does the API consume the `description` field at all, or is it purely for the human preparing this file?

metadeploy/adminapi/translations.py
```diff
@@ -0,0 +1,54 @@
...
+    for model in TranslatableModel.__subclasses__():
+        if not hasattr(model, "get_translation_strategy"):
+            continue
+        for obj in model.objects.all():
+            update_translations(obj, [lang])
+
+
+def update_translations(obj, langs=None):
+    """Update one object's translations from the Translation model"""
+    assert obj.get_current_language() == "en-us"
```
davidmreed: Does this assertion imply that a customer could not run MetaDeploy with a primary language other than English and utilize translations?

metadeploy/adminapi/api.py
davisagli: It's context for the human translators.

metadeploy/adminapi/translations.py
davisagli: Good point; I should change this to use the first language from settings.

PR #1170, opened 2020-04-06
Week 15 dependency updates

Br4nd0R:


requirements/base.txt
```diff
@@ -1,16 +1,16 @@
-cumulusci==3.9.1
+cumulusci==3.10.1
```
davisagli, line 1:
```suggestion
cumulusci==3.10.0
```

There is no 3.10.1

PR #1172, opened 2020-04-07
Better error message when API not enabled in org

Br4nd0R:
* We now display a better error message when we detect that login fails with a 403 status code, and the `errorCode` confirms that the org does not have its REST API enabled


metadeploy/multisalesforce/views.py
```diff
@@ -55,7 +56,16 @@ def get_org_details(self, extra_data, token):
...
+            resp.status_code == 403
+            and json.loads(resp.text)[0]["errorCode"] == "API_DISABLED_FOR_ORG"
+        ):
+            raise SalesforcePermissionsError(
+                "Sorry, MetaDeploy requires an org with the API enabled."
+            )
+        else:
+            resp.raise_for_status()
+
         return resp.json()
```
davisagli, line 69: A small tweak to parse the json upfront regardless of whether we're handling an error or not, and to be more defensive in case the response isn't in the expected format.

```suggestion
        resp_json = resp.json()
        if (
            resp.status_code == 403
            and len(resp_json) and resp_json[0]["errorCode"] == "API_DISABLED_FOR_ORG"
        ):
            raise SalesforcePermissionsError(
                "Sorry, MetaDeploy requires an org with the API enabled."
            )
        else:
            resp.raise_for_status()
        return resp_json
```

PR #1202, opened 2020-04-13
Week 16 dependency updates

Br4nd0R:


davisagli:
@Br4nd0R I ran into this error last week -- we need to hold html-webpack-plugin back to major version 3 until they either re-add a feature they removed or we have time to figure out what to do instead.

davisagli:
@Br4nd0R the build failed due to presence of both yarn.lock and package-lock.json. The latter should not have been added.

Br4nd0R:
Just saw that. Removing on another PR

PR #1330, opened 2020-05-21
🛢 First steps toward ability to publish revised plans

davisagli:
1. Add a commit_ish field to Plan which can optionally be used to override the Version's commit_ish
2. Remove the constraint to only allow one plan per version-plantemplate.
2. Add a created_at field to Plan which will be used to pick the most recent plan when a version-plantemplate has more than one. There is a migration which tries to populate this based on commit dates from github. The field is not actually used yet because I want to make sure the migration worked properly first.


metadeploy/api/migrations/0087_plan_revisions.py
```diff
@@ -0,0 +1,73 @@
...
+from metadeploy.api.cci_configs import extract_user_and_repo
+
+DEFAULT_TIMESTAMP = parse_datetime("2019-01-01T00:00:00Z")
+
+
+def populate_created_at(apps, schema_editor):
+    model = apps.get_model("api", "Plan")
+
+    repos = {}
+    gh = github3.login(token=settings.GITHUB_TOKEN)
```
jlantz, line 17: Is there any risk in calling out to the GitHub API as part of a migration script?

davisagli: @jlantz it does mean the token needs to be set while running the migration. I should move this line inside the try/except so that we fall back to the default in that case rather than breaking.

davisagli: I suppose there's also a risk of hitting the github API rate limit, but I doubt we have enough plans for that to be a problem at this point.

PR #1483, opened 2020-06-30
📦 🐈 week 27 dependency updates

davisagli:


davisagli:
@boakley it's a change to comply with a slightly stricter type checking rule in the newer version of Flow

PR #1541, opened 2020-07-21
Allow use of GitHub App

jlantz:
Allows for either `GITHUB_TOKEN` to be used with a personal access token or `GITHUB_APP_ID` and `GITHUB_APP_KEY` for using a GitHub App.

- Removed the setting of the github service in the constructed keychain since CumulusCI now will use the `GITHUB_*` environment variables rather than fetching the service from the keychain if the environment variables are set

NOTE: This won't work for any of the github_* tasks in CumulusCI as BaseGithubTask requires that the github service is set in the keychain.  This should be addressed in CumulusCI to have a more intelligent check that checks if the GITHUB_* environment variables are set and falls back to checking the keychain if not.  However, that change isn't required for this PR since MetaDeploy plans typically don't use the github_* tasks from CumulusCI.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -312,7 +312,24 @@ STATIC_ROOT = str(PROJECT_ROOT / "staticfiles")
 CONNECTED_APP_CLIENT_SECRET = env("CONNECTED_APP_CLIENT_SECRET")
 CONNECTED_APP_CALLBACK_URL = env("CONNECTED_APP_CALLBACK_URL")
 CONNECTED_APP_CLIENT_ID = env("CONNECTED_APP_CLIENT_ID")
-GITHUB_TOKEN = env("GITHUB_TOKEN")
+GITHUB_TOKEN = env("GITHUB_TOKEN", default=None)
+if GITHUB_TOKEN in ("", "None"):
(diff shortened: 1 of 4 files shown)


config/settings/base.py
```diff
@@ -312,7 +312,24 @@ def env(name, default=NoDefaultValue, type_=str):
...
-GITHUB_TOKEN = env("GITHUB_TOKEN")
+GITHUB_TOKEN = env("GITHUB_TOKEN", default=None)
+if GITHUB_TOKEN in ("", "None"):
+    GITHUB_TOKEN = None
+GITHUB_APP_ID = env("GITHUB_APP_ID", default=None)
+if GITHUB_APP_ID in ("", "None"):
+    GITHUB_APP_ID = None
+GITHUB_APP_KEY = env("GITHUB_APP_KEY", default=None)
+if GITHUB_APP_KEY in ("", "None"):
+    GITHUB_APP_KEY = None
```
davisagli, line 323: Is there a particular reason we need to support "None" as a literal value for these? If not I would be inclined to skip this normalization and just check if the vars are truthy below (empty string is not).

PR #1627, opened 2020-08-18
Feature/fix preflight translation display

davidmreed:
This PR fixes an issue manifesting as preflight result messages having the English word "encountered" embedded inside translated text. The source issue is that a dynamically constructed translatable string name was partially translated _before_ constructing.


davisagli:
Oddbird fixed this in https://github.com/oddbird/MetaDeploy/pull/144

PR #1678, opened 2020-08-28
📦 Add support for storing images in the db if AWS isn't configured

davisagli:
If the settings related to AWS are not configured, metadeploy will use django-binary-database-files to store images in the postgres db.

Draft pending third-party software approval of the new dependency.


jlantz:
3pp is approved, we should be able to merge this now

jlantz:
We should look into adding the File model to the Django Admin

jlantz (approved):

jlantz (approved):

PR #1875, opened 2020-10-15
Capitalize the recommended string for MetaDeploy UI

shibinkunnath:
Capitalize the recommended string for MetaDeploy UI to look nicer in the UI.

Discussed this with @jlantz yesterday. From the code search, it is used only in one file for which I believe it is ok to capitalize. But please check and let me know if we're good to go ahead with this change.

Current screenshot where it is not capital
Screen Shot 2020-10-14 at 12 36 54 PM


davisagli:
@shibinkunnath This looks good; thanks for the PR. The builds which are required to merge only run for PRs created from a branch within the SFDO-Tooling organization, so I'll submit a new PR with the same changes.

shibinkunnath:
@davisagli Ok sure, thanks

davisagli (approved):

PR #1978, opened 2020-11-07
Hotfix translations and dropdown menu positions.

jgerigmeyer:
- consistent capitalization in steps table
- fix plan/product not-allowed translations
- fix plan detail CTA login btn dropdown overflowing screen


locales/en/translation.json
```diff
@@ -198,12 +197,10 @@
   "productWithVersion": "<0>{{product}}</0>, <2>{{version}}</2>",
   "recommended": "Recommended",
   "reload the page.": "reload the page.",
-  "required": "required",
-  "skipped": "skipped",
```
jgerigmeyer, line 202: @davisagli These are no longer used. They could be removed from all translation files, if you'd like.

locales_dev/en/translation.json
```diff
@@ -103,9 +102,9 @@
   "We couldn’t find any products. Try again later?": "We couldn’t find any products. Try again later?",
   "Who can access this shared link?": "Who can access this shared link?",
   "You are in offline mode. We are trying to reconnect, but you may need to": "You are in offline mode. We are trying to reconnect, but you may need to",
-  "a_or_b": "<0></0> or <2></2>",
+  "a_or_b": "<0><0></0> or </0><1></1>",
```
davisagli, line 105: Hmm. The reason I created this translation key was to make it obvious that it's a translation of an alternative between two choices, and to make it easy to put things in a different order if there's a language where the "or" doesn't go between the two choices (maybe that's only hypothetical). Maybe we don't have a good alternative but this makes it less obvious what's going on.

src/js/components/plans/ctaButton.tsx
```diff
@@ -67,7 +67,7 @@ export const LoginBtn = ({
     triggerClassName="slds-size_full"
     label={label}
     menuPosition="relative"
-    nubbinPosition="top"
+    flipped
```
jgerigmeyer, line 70: I recognize this isn't as important with the full-width btns in production, but I think it still doesn't hurt, and it becomes necessary with the change to smaller btns in the OddBird fork.

davisagli (commented):
Thanks for these fixes. They mostly make sense but I wish we didn't have to handle the "or" translation this way.

davisagli (approved):
I'll go ahead and merge this and if we find a better way to handle the "or" translation we can make that fix separately.

PR #2019, opened 2020-11-16
🛢 Store preflight logs

davisagli:
Occasionally we get requests to investigate why preflights failed, but their logs are currently not stored. This fixes that by streaming logs to a buffer which then gets written to the PreflightResult model once the preflight flow is done (similar to how installation job logs are handled).


Br4nd0R (approved):
looks good!

PR #2184, opened 2021-01-06
fix jwt auth for connected orgs

davisagli:
cci needs an OrgConfig with a username set in order to complete JWT auth for a connected org. (And for other things, but that's what revealed that it's missing.) This is a fairly recent regression from the work on the scratch org feature, and hasn't affected production because we haven't set SFDX_HUB_KEY there yet, so cci is still using the refresh token flow there.

### DIFF
--- a/metadeploy/api/jobs.py
+++ b/metadeploy/api/jobs.py
@@ -135,13 +135,7 @@ def run_flows(*, plan, skip_steps, result_class, result_id):
     """
     result = result_class.objects.get(pk=result_id)
     scratch_org = None
-    token = None
-    token_secret = None
-    instance_url = None
-    if result.user:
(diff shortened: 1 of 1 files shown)


davidmreed (approved):

jgerigmeyer:
@davisagli We're using OrgConfig in two places in Metecho; I'm guessing not, but want to be sure this fix doesn't need to be applied there too?

- https://github.com/oddbird/Metecho/blob/main/metecho/api/sf_run_flow.py#L99
- https://github.com/oddbird/Metecho/blob/main/metecho/api/sf_run_flow.py#L117

davisagli:
@jgerigmeyer since auth in Metecho is working, `username` is presumably already in the `config` dict there.

PR #2206, opened 2021-01-11
Week 2 updates

Br4nd0R:


package.json
```diff
@@ -90,38 +90,38 @@
     "@types/classnames": "^2.2.11",
     "@types/core-js": "^2.5.4",
     "@types/js-cookie": "^2.2.6",
-    "@types/lodash": "^4.14.166",
+    "@types/lodash": "^4.14.167",
     "@types/react": "^16.14.2",
     "@types/react-document-title": "^2.0.4",
-    "@types/react-dom": "^16.9.10",
-    "@types/react-redux": "^7.1.14",
+    "@types/react-dom": "^17.0.0",
```
davisagli, line 96: @Br4nd0R We're still using major version 16 of React, so we should avoid updating other react packages to ^17

davisagli (changes requested):

davisagli (approved):

PR #2232, opened 2021-01-15
Warn Checks Should Not Block Error Checks

Br4nd0R:
# Changes
* `PreflightFlowCallback.post_flow` now saves all results for a given step to the results of the PreflightResult or Job.

## Outstanding
- [x] Figure out front-end logic for displaying pre-flight results when more than one is present. Should we only show one based on precedence (i.e. errors first, then warnings), or should we display all results at the same time.
- [x] Need to find other potential places where changing the structure of `PreflightResult.results` will break things.
- [x] Need to decide whether to alter the current format of how `Job.results` are captured (i.e. have a potential list of error associated with a given step num) or if the `ErrorWarningCountMixin` should process counting errors differently between `Jobs` and `PreflightResult`s. Or maybe the mixin need to be dissolved if the format of results is going to differ between the two objects. Question: Is it possible for a single step to have more than one exception? If not, then we likely want to dissolve the mixing and update each model accordingly.

## QA Scenarios Checked
- [x] When warning check is triggered it _does not_ block subsequent error checks from being triggered.
Sample Preflight Check:
```
[{"when": "'.star-wars.' not in org_config.instance_url", "action": "warn", "message": "star-wars is not in org_config.instance_url."}, {"when": "'.trogdor.' not in org_config.instance_url", "action": "error", "message": "Trogdor must burninate the org_config.instance_url"}]
```
- [x] When only warnings are present on preflight results, they are displayed by `<PlanDetail>`.
- [x] When only errors are present on preflight results, they are displayed by `<PlanDetail>`.
- [x] When both warnings _and_ errors are present on preflight results, only errors are displayed by `<PlanDetail>`.
- [x] When step-level warnings are present they trigger the `<PreflightWarningModal>` to be shown by the `<ctaButton>`.
- [x] When only plan-level warnings are present, no `<PreflightWarningModal>` is presented to the user. (i.e. they are displayed to the user in `<PlanDetail>`
- [x] Successful full install of EDA
- [x] Ensured that active steps display correctly in the `StepTable`


src/js/components/jobs/shareModal.tsx
```diff
@@ -83,17 +83,20 @@ class ShareModal extends React.Component<WrappedProps> {
...
-            result.status === CONSTANTS.RESULT_STATUS.ERROR &&
-            result.message
-          ) {
-            const step = plan.steps.find((s) => s.id === id);
-            if (step) {
-              stepError = result.message;
-              stepName = step.name;
-              break;
+        // will return the last error present in the list
+        // of results for a given step
```
davisagli, line 87: Looks like it's actually returning the first error

src/js/components/plans/detail.tsx
```diff
@@ -216,8 +216,14 @@ class PlanDetail extends React.Component<Props, State> {
     }
     for (const step of plan.steps) {
       const { id } = step;
-      const result = preflight?.results?.[id];
-      if (!result || result.status !== RESULT_STATUS.HIDE) {
+      const results = preflight?.results?.[id];
+      if (results) {
+        for (const result of results) {
+          if (!result || result.status !== RESULT_STATUS.HIDE) {
+            steps.push(step);
```
davisagli, line 223: Won't this add a step to the list multiple times if it has multiple non-HIDE results?

src/js/components/plans/detail.tsx
```diff
@@ -235,12 +241,14 @@ class PlanDetail extends React.Component<Props, State> {
...
-      if (result) {
-        hidden = result.status === RESULT_STATUS.HIDE ? result : null;
-        skipped = result.status === RESULT_STATUS.SKIP ? result : null;
-        optional = result.status === RESULT_STATUS.OPTIONAL ? result : null;
+      if (results) {
+        for (const result of results) {
+          hidden = result.status === RESULT_STATUS.HIDE ? result : null;
+          skipped = result.status === RESULT_STATUS.SKIP ? result : null;
+          optional = result.status === RESULT_STATUS.OPTIONAL ? result : null;
+        }
```
davisagli, line 251: This is overwriting these flags on each iteration over the results. Instead I think we want to _collect_ them. i.e. `hidden` should be true if any result was HIDE, `skipped` should be true if any result was SKIP, etc

src/js/components/plans/stepsTable/index.tsx
```diff
@@ -88,16 +88,27 @@ class StepsTable extends React.Component<Props, State> {
...
+  isActiveStep = (stepId: string, job: Job) => {
+    if (job.results[stepId]) {
+      for (const result of job.results[stepId]) {
+        if (!result.status) {
+          return true;
+        }
+      }
+    }
+    return false;
+  };
```
davisagli, line 110: I think you've changed the logic here in a way that doesn't make sense. The point was to find the first step that doesn't have any results yet. But this will find the first step that has results where one of the results doesn't have a status (is that even possible?) Probably worth QA to make sure that wherever this is used is getting the right step.

src/js/components/plans/stepsTable/index.tsx
```diff
@@ -109,15 +120,27 @@ class StepsTable extends React.Component<Props, State> {
...
+  stepHasLogs = (stepId: string, job: Job) => {
+    if (job.results[stepId]) {
+      for (const result of job.results[stepId]) {
+        if (result.logs) {
+          return true;
+        }
+      }
+    }
+    return false;
+  };
```
davisagli, line 142: Since this is the main job, not the preflight, I think you can safely assume that there will be either 0 results or 1. So you could just check job.results[stepId][0].logs rather than looping

src/js/components/plans/stepsTable/installDataCell.tsx
```diff
@@ -180,11 +182,13 @@ class PreflightCell extends React.Component<DataCellProps> {
...
-    const result = preflight?.results?.[id];
+    const results = preflight?.results?.[id];
     let skipped, optional, content;
-    if (result) {
-      skipped = result.status === RESULT_STATUS.SKIP ? result : null;
-      optional = result.status === RESULT_STATUS.OPTIONAL ? result : null;
+    if (results) {
+      for (const result of results) {
+        skipped = result.status === RESULT_STATUS.SKIP ? result : null;
+        optional = result.status === RESULT_STATUS.OPTIONAL ? result : null;
```
davisagli, line 190: Again, I think we want to collect values from multiple results here, not overwrite with the values based on the last result.

src/js/components/plans/stepsTable/requiredDataCell.tsx
```diff
@@ -24,11 +24,13 @@ const RequiredDataCell = (props: DataCellProps) => {
...
-  const result = preflight?.results?.[id];
+  const results = preflight?.results?.[id];
   let skipped, optional;
-  if (result) {
-    skipped = result.status === RESULT_STATUS.SKIP ? result : null;
-    optional = result.status === RESULT_STATUS.OPTIONAL ? result : null;
+  if (results) {
+    for (const result of results) {
+      skipped = result.status === RESULT_STATUS.SKIP ? result : null;
+      optional = result.status === RESULT_STATUS.OPTIONAL ? result : null;
```
davisagli, line 32: Ditto

src/js/store/plans/reducer.ts
```diff
@@ -33,16 +33,18 @@ export type StepResult = {
   message?: string;
   logs?: string;
 };
-export type PreflightErrors = {
-  [key: string]: StepResult;
+
+export type JobErrors = {
+  [key: string]: StepResult[];
 };
```
davisagli, line 39: Maybe this type should be called PlanResults, since it's used for both Jobs and Preflights, and a single result might express a status other than error

test/js/components/plans/preflightResults.test.js
```diff
@@ -101,7 +101,7 @@ describe('<PreflightResults />', () => {
   describe('completed preflight with errors and warnings', () => {
     test('displays error/warning messages', () => {
       const { getByText } = setup();
-
+      // TODO: Only errors should be displayed
```
davisagli, line 104: todo

davisagli (changes requested):
Good work -- overall looks good, but I have questions about a few details

src/js/components/plans/stepsTable/index.tsx
Br4nd0R: @davisagli - I agree something is off here, but isn't the original logic similar? The `if` check will be true when either:
(1) There are no results for the step yet (and `?.` will return undefined)
(2) There are results but a `status` is not present (i.e. the only "results" are the logs set by the `ResultSpoolLogger`.

If I make the function return `true` if there are no results for a step, then it looks like the "in progress" step is one ahead of where it should be. Put another way, the step that has logs currently being output is not shown as the "in progress" step. So I'm guessing this is why the function was originally checking for the presence of `status`, as it seems to be set by `JobFlowCallback.post_task()`.

Br4nd0R: Ok I think I got it working :)

src/js/components/plans/detail.tsx
Br4nd0R: @davisagli - Do we know if it's safe to assume that there will ever only be one result with a "status" value? I'm trying to determine wether or not `hidden`, `skipped`, and `optional` need to be lists of results or if it will suffice to just use the first results found with a given status.

davisagli: @Br4nd0R A step with multiple preflight checks can have multiple results, each with its own status. e.g. [{"status": "skip", "message": "Skipped because X"}, {"status": "hide"}]. The goal here is to end up with a boolean for hidden, skipped, and optional indicating whether any of the results had that status.

So I think something like this would do it, no loop needed:
```
const results = preflight?.results?.[id] || [];
const hidden = results.some(result => result.status === RESULT_STATUS.HIDE);
const skipped = results.some(result => result.status === RESULT_STATUS.SKIP);
const optional = results.some(result => result.status === RESULT_STATUS.OPTIONAL);
```

src/js/components/plans/stepsTable/index.tsx
davisagli: Makes sense. I forgot about the logs when I wrote this comment.

davisagli:
@Br4nd0R there are a few Python tests for ResultSpoolLogger that need to be adjusted

src/js/components/plans/detail.tsx
Br4nd0R: `installDataCell.tsx` will need to have an actual `result` object captured. It later references `skipped?.label` to pass a label to the `<Checkbox>` component. Would you want me to aggregate all "messages" from results with a status of "skip" and pass that to the `<Checkbox>` component? Should we expect multiple "skip" results for a single step?

davisagli: @Br4nd0R ah ok, I kinda thought it must be something like that but I wasn't spotting where the `skipped` result object was used. It's theoretically possible that there could be multiple checks which each set a status of `skip` with a different message -- but it's enough of an edge case that I'm fine calling that out of scope for this particular change and only show the first one.

davisagli:
@Br4nd0R super close but frontend coverage is no longer at 100%

davisagli (approved):

PR #2257, opened 2021-01-21
Only use the generic Install {product} {version} label when the default value ends with a number

davisagli:
This fixes it so that we use this label in the same cases where the metadeploy_publish task generates it. Before, a label like "Install foo" would be left alone by metadeploy_publish, but MetaDeploy would still convert it to "Install {product} {version}" and look for a translation that doesn't exist.

### DIFF
--- a/metadeploy/adminapi/translations.py
+++ b/metadeploy/adminapi/translations.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


shibinkunnath (approved):
Thanks David. Looks good to me.

Br4nd0R (approved):
LGTM

PR #2341, opened 2021-02-12
Gracefully handle ip restriction errors when logging in

Br4nd0R:
# Changes
* MetaDeploy now has `custom_500_view` which has default messages for both generic 500 errors as well as when MetaDeploy detects that IP Login Ranges are in place for a user. The message displayed for IP Login Range errors can be overridden by setting the `IP_RESTRICTED_MESSAGE` config variable in Heroku.

### DIFF
--- a/metadeploy/tests/views.py
+++ b/metadeploy/tests/views.py
@@ -46,10 +46,10 @@ def test_custom_500_view__ip_restricted_error(render):
         )
     except OAuth2Error:
         allow_list = "0.0.0.1, 0.0.0.2, 0.0.0.3"
-        with mock.patch("metadeploy.views.IPS_TO_ALLOWLIST", allow_list):
+        with mock.patch("metadeploy.views.IP_RESTRICTED_MESSAGE", allow_list):
             factory = RequestFactory()
             request = factory.get("/accounts/salesforce/login/callback/")
             request.user = AnonymousUser()
             custom_500_view(request)
 
-    assert allow_list in render.call_args[1]["context"]["JS_CONTEXT"]["error_message"]
+    assert allow_list == render.call_args[1]["context"]["JS_CONTEXT"]["error_message"]
--- a/metadeploy/tests/views.py
+++ b/metadeploy/tests/views.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 from unittest import mock
 
@@ -45,12 +46,11 @@ def test_custom_500_view__ip_restricted_error(render):
             'Error retrieving access token: b\'{"error":"invalid_grant","error_description":"ip restricted"}\''
         )
     except OAuth2Error:
-        factory = RequestFactory()
-        request = factory.get("/accounts/salesforce/login/callback/")
-        request.user = AnonymousUser()
-        custom_500_view(request)
+        test_ips = "0.0.0.1, 0.0.0.2, 0.0.0.3"
+        with mock.patch.dict(os.environ, {"IPS_TO_WHITELIST": test_ips}):
+            factory = RequestFactory()
+            request = factory.get("/accounts/salesforce/login/callback/")
+            request.user = AnonymousUser()
+            custom_500_view(request)
 
-    assert (
-        "We've detected that your org has ip login recstrictions in place."
-        in render.call_args[1]["context"]["JS_CONTEXT"]["error_message"]
-    )
+    assert test_ips in render.call_args[1]["context"]["JS_CONTEXT"]["error_message"]
--- a/metadeploy/tests/views.py
+++ b/metadeploy/tests/views.py
@@ -1,16 +1,18 @@
+import pytest
 from unittest import mock
 
-import pytest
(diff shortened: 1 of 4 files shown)


metadeploy/tests/views.py
```diff
@@ -1,16 +1,18 @@
...
 
-from ..views import custom_permission_denied_view
+from ..views import custom_permission_denied_view, custom_500_view
 
 
 @pytest.mark.django_db
 @mock.patch("metadeploy.views.render")
 def test_custom_permission_denied_view__sf_permissions(render):
-    request = RequestFactory()
+    request = RequestFactory().get("path")
```
davisagli, line 15: What's this for?

metadeploy/views.py
```diff
@@ -1,16 +1,38 @@
...
     )
+
+
+def custom_500_view(request):
+    message = GENERIC_ERROR_MSG
+    error_type, value, traceback = sys.exc_info()
+
+    if "ip restricted" in value.args[0]:
+        message = (
+            "We've detected that your org has ip login recstrictions in place. "
```
davisagli, line 29: typo

davisagli: I think the actual name of the feature is Login IP Ranges https://developer.salesforce.com/docs/atlas.en-us.securityImplGuide.meta/securityImplGuide/users_profiles_epui_login_ip_ranges_edit.htm

davisagli: Also it'd be more accurate to say user rather than org. Login IP Ranges are added to a profile, and different users can have different profiles

metadeploy/views.py
```diff
@@ -1,16 +1,38 @@
...
+            "Please whitelist the IP addresses listed in the plan's description."
```
davisagli, line 30: What if they haven't been added to the plan description? I'm not sure we can provide generic instructions for how to resolve this for all instances of metadeploy.

metadeploy/views.py
```diff
@@ -1,16 +1,38 @@
...
     return render(
         request,
         "index.html",
         context={"JS_CONTEXT": {"error_message": message}},
         status=403,
     )
+
+
+def custom_500_view(request):
+    message = GENERIC_ERROR_MSG
```
davisagli, line 24: Have you confirmed that the actual error message still gets logged? I suspect it does, but we don't want the nicer error view to get in the way of us being able to look up what happened, so worth double checking.

davisagli (changes requested):

metadeploy/tests/views.py
Br4nd0R: RequestFactory() returns a factory object, not an actual request. So these tests weren't setup properly by me in the first place.

metadeploy/views.py
Br4nd0R: fixed with a new setting so we can list IPs to whitelist in a config var.

metadeploy/views.py
Br4nd0R: Confirmed that the error message raised is still found in the app logs (along with the stacktrace).

metadeploy/tests/views.py
davisagli: Oh yeah, I see.

metadeploy/views.py
davisagli: @Br4nd0R oh, that's a good idea. But we should only include the sentence that references IPS_TO_WHITELIST if it has actually been configured.

davisagli: @Br4nd0R and it should be allowlist, not whitelist

Br4nd0R: @davisagli - Are you saying the at the message should say, "Please allowlist the IP addressess..." or that the `IPS_TO_WHITELIST` var should be `IPS_TO_ALLOWLIST`?

davisagli: @Br4nd0R we should not use "whitelist" anywhere; it's non-inclusive language.

I think the message should say "Unable to access this org because your user has IP Login Ranges that block access." and if IPS_TO_ALLOWLIST is set, add "Please ensure that the following IP addresses are included in the IP Login Ranges for your user's Profile:"

metadeploy/views.py
```diff
@@ -1,16 +1,45 @@
...
+    value = sys.exc_info()[1]
+
+    if "ip restricted" in value.args[0]:
+        message = IP_RESTRICTED_MSG
+        if IPS_TO_ALLOWLIST:
+            message += (
+                " Please ensure that the following IP addresses are "
+                f"included in the IP Login Ranges for you user's Profile: {IPS_TO_ALLOWLIST}"
+            )
```
davisagli, line 38: @Br4nd0R Now I'm thinking we should make the entire IP_RESTRICTED_MSG configurable in a setting (and take out IPS_TO_ALLOWLIST). It'll keep things flexible if a particular instance wants to use different wording. Default would be "Unable to access this org because your user has IP Login Ranges that block access." like you have above.

davisagli (approved):

PR #2403, opened 2021-02-24
Fix exception handler: don't assume exception has a response

davisagli:
This fixes a misleading error in the chain of stack traces that I noticed while looking into another problem.


Br4nd0R (dismissed):

jstvz:
@Br4nd0R Any idea whether this is still a valid change?

Br4nd0R:
@jstvz - I think this just fell through the cracks. Should be good to merge.

PR #2503, opened 2021-03-18
Fix: jobs with hidden steps fail

Br4nd0R:


metadeploy/api/serializers.py
```diff
@@ -529,9 +529,10 @@ def _validate_results(self, data):
             # results are read-only except during creation
             del data["results"]
         else:
-            # make sure results can't be set initially except to hide steps
-            if any(result["status"] != HIDE for result in data["results"].values()):
-                raise serializers.ValidationError(_("Invalid initial results."))
+            for result in data["results"]:
+                # make sure results can't be set initially except to hide steps
+                if any(r["status"] != HIDE for r in result.values()):
+                    raise serializers.ValidationError(_("Invalid initial results."))
```
davisagli, line 535: This doesn't look right. The top level of data["results"] is a dict mapping step id -> result object. Iterating like this will give you the keys (step ids) so `result.values()` won't work below. I think you want:

```suggestion
            for step_id, results in data["results"].items():
                # make sure results can't be set initially except to hide steps
                if any(result["status"] != HIDE for result in results):
                    raise serializers.ValidationError(_("Invalid initial results."))
```

davisagli (changes requested):

davisagli (approved):

PR #2536, opened 2021-03-30
Treat skipped steps as optional for validity

davidmreed:
When a MetaDeploy plan is run that includes preflight checks with the `skip` action, and the conditions are met to skip that step, MetaDeploy presents an error:

> Invalid steps for plan

This PR aims to fix the issue by correcting the identification of optional steps to include those in `SKIP` status.


davisagli (approved):
The fix makes sense to me.

davisagli:
@davidmreed I'm not sure why the builds didn't get triggered. Could you try pushing a new commit and see if that helps?

davisagli (approved):

PR #2568, opened 2021-04-01
Recommend github app instead of token

davisagli:
It's a best practice to have MetaDeploy authenticate to github as a GitHub app rather than using a personal access token.

- Update docs and defaults to show setting GITHUB_APP_ID and GITHUB_APP_KEY instead of GITHUB_TOKEN
- Send the release phase event to GitHub using a Python script which support authenticating using the app credentials.


Br4nd0R (approved):
lgtm!

PR #2674, opened 2021-04-29
📦 🐈 dependency updates

Br4nd0R:


package.json
```diff
@@ -76,18 +76,18 @@
...
+    "@babel/core": "^7.13.16",
     "@babel/plugin-proposal-class-properties": "^7.12.13",
     "@babel/plugin-proposal-export-default-from": "^7.12.13",
     "@babel/plugin-proposal-export-namespace-from": "^7.12.13",
     "@babel/plugin-proposal-object-rest-spread": "^7.12.13",
     "@babel/preset-env": "^7.13.15",
     "@babel/preset-react": "^7.13.13",
     "@babel/preset-typescript": "^7.13.0",
-    "@testing-library/jest-dom": "^5.11.10",
+    "@testing-library/jest-dom": "^5.12.0",
```
davisagli, line 87: There are some errors from jest in the "lint" build.

requirements/test.txt
```diff
@@ -3,23 +3,23 @@
 -r base.txt
 
 Sphinx==3.5.4
-black==20.8b1
+black==21.4b1
```
davisagli, line 6: Need to make sure we've run the new version on existing files.

davisagli (changes requested):

package.json
Br4nd0R: @davisagli - I tried playing around to fix this error which is popping up in the tests but I'm not sure of how to change the syntax based on what `eslint-plugin-jest` shows for [this particular error](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md).

The `expect()` calls can't be in the call to `catch()` because `catch()` only executes conditionally (if there is a rejected case for the promise). Not sure how to shift things around to have `expect()` calls not in the `catch()` method. 

If you can show me an example of how to do it for one, I can fix the rest of them.

davisagli: @Br4nd0R It's a lint warning (about a situation that could lead to silently skipping test assertions) rather than a hard failure, so I just disabled the lint rule.

davisagli (approved):

PR #3028, opened 2021-08-04
Cron Job for calculating average plan duration

Br4nd0R:
# Changes
* MetaDeploy now calculates the median plan runtime and caches this value on the Plan model.


config/settings/base.py
```diff
@@ -464,6 +464,10 @@ def env(name, default=NoDefaultValue, type_=str):
         "func": "metadeploy.api.jobs.expire_preflights_job",
         "cron_string": "* * * * *",
     },
+    "calculate_average_plan_runtimes": {
+        "func": "metadeploy.api.jobs.calculate_average_plan_runtime_job",
+        "cron_string": "0 0 * * *" # run dialy at midnight
```
davisagli, line 469:
```suggestion
        "cron_string": "0 0 * * *" # run daily at midnight
```

metadeploy/api/models.py
```diff
@@ -501,6 +501,13 @@ class Plan(HashIdMixin, SlugMixin, AllowedListAccessMixin, TranslatableModel):
         help_text="Lifetime of Scratch Orgs created for this plan. Will inherit the "
         "global default value if left blank.",
     )
+    calculated_average_duration = models.IntegerField(
+        "Average duration of a plan (seconds)",
+        null=True,
+        blank=True,
+        validators=[MinValueValidator(0)],
+        help_text="The duration between the enqueing of a job and its successful completion."
```
davisagli, line 509:
```suggestion
        help_text="The duration between the enqueueing of a job and its successful completion."
```

src/js/store/plans/reducer.ts
```diff
@@ -21,7 +21,7 @@ export type Plan = {
   is_listed: boolean;
   is_allowed: boolean;
   not_allowed_instructions: string | null;
-  average_duration: string | null;
+  calculated_average_duration: string | null;
```
davisagli, line 24: Rather than changing the name that is serialized, I think you want to change `average_duration` on the PlanSerializer to be a SerializerMethodField so that you can implement the logic for reading it as a method instead of getting it directly from the field with the same name.

davisagli (changes requested):
Mostly looks good so far but I think we can avoid needing to change the frontend.

metadeploy/adminapi/tests/api.py
```diff
@@ -60,6 +60,7 @@ def test_list(self, admin_api_client, plan_factory):
         assert json == {
             "data": [
                 {
+                    "calculated_average_duration": None,
```
Br4nd0R, line 63: @davisagli - I'm trying to figure out why these were needed. The `PlanViewSet` specifies the `PlanSerializer` as the serializer class. That serializer doesn't reference calculated_average_duration at all, so why is this field showing in these view sets?

davisagli: @Br4nd0R This took me a second to figure out. The PlanSerializer in metadeploy/api/serializers.py specifies `fields` and only includes those fields. But this is a test for the admin API which uses a different PlanSerializer in https://github.com/SFDO-Tooling/MetaDeploy/blob/2b2f222dc82e43f6d233136837f9b0e23ed672c3/metadeploy/adminapi/api.py, and that one includes all fields. You can add `exclude` there if we don't want it https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include

Br4nd0R: We can leave it in; I don't think it's hurting anything.

davisagli: @Br4nd0R it does mean clients of the admin API can write a value to this field, which feels a bit wrong.

metadeploy/api/migrations/0114_plan_calculated_average_duration.py
```diff
@@ -0,0 +1,22 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="plan",
+            name="calculated_average_duration",
+            field=models.IntegerField(
+                blank=True,
+                null=True,
+                verbose_name="Average duration of a plan (seconds)",
```
davisagli, line 19: I just noticed that we're calling it the average but calculating it as the median (which is what we want)

metadeploy/api/serializers.py
```diff
@@ -223,6 +224,11 @@ def get_not_allowed_instructions(self, obj):
     def get_requires_preflight(self, obj):
         return obj.requires_preflight
 
+    def get_average_duration(self, obj):
+        """Plan.average_duration is an expensive query,
+        so we prefer the already calculated value if available."""
+        return obj.calculated_average_duration or obj.average_duration
```
davisagli, line 230:
```suggestion
        return obj.calculated_average_duration
```

so that we don't always try to recalculate when calculated_average_duration is None because not enough jobs have run yet

davisagli (changes requested):

davisagli (approved):

PR #3059, opened 2021-08-24
Updates to change traffic control check

davisagli:
1. refactoring
2. actually enforce the check
3. don't run the db migration until we've confirmed there is an approved change case


codexmarcel (approved):
LGTM per live review.

PR #3160, opened 2021-09-20
🐈  JS Dependency Updates

Br4nd0R:


Br4nd0R:
@davisagli - I pinned `node-fetch` to mitigate this issue; however, this only fixes one of the dependabot alerts. I'm not sure how to fix the issues where one package is pinned by another transient one. [Some of them](https://github.com/SFDO-Tooling/MetaDeploy/security/dependabot/yarn.lock/browserslist/open) are referenced by many other packages, so would we need to pin each of those to a version that doesn't use the vuln version of the dependency in question?

davisagli:
Replaced by https://github.com/SFDO-Tooling/MetaDeploy/pull/3161

PR #3161, opened 2021-09-20
🐈 Upgrade JS dependencies.

jgerigmeyer:
Upgrades JS dependencies, addressing security alerts.

Holding off on upgrading i18next for now; there are some [breaking changes in v21](https://www.i18next.com/misc/migration-guide#v-20-x-x-to-v-21-0-0) that don't (currently) play nicely with our auto-generated translation file. This may not be a problem in production, but will make it more difficult to track changes in translation strings going forward.


jgerigmeyer:
@davisagli I can't see the Dependabot security alerts (even on the OddBird fork with them enabled), but I _think_ this should resolve the ones I'm aware of.

davisagli (approved):

PR #3205, opened 2021-10-06
Job and preflight completion logging

davidmreed:
Log when jobs and preflight checks complete with various statuses for ingestion into metrics.

Partially dependent on W-9935537 (for new fields).

### DIFF
--- a/metadeploy/api/jobs.py
+++ b/metadeploy/api/jobs.py
@@ -66,7 +66,7 @@ class JobLogStatus(str, enum.Enum):
 
 
 @contextlib.contextmanager
-def finalize_result(plan: Plan, result: Union[Job, PreflightResult]):
+def finalize_result(result: Union[Job, PreflightResult]):
     start_time = timezone.now()
     end_time = None
     log_status = None
@@ -130,9 +130,17 @@ def finalize_result(plan: Plan, result: Union[Job, PreflightResult]):
         elif isinstance(result, Job):
             job_type = JobType.JOB
 
-        context = f"{plan.version.product.title} {plan.version.label}"
+        context = f"{result.plan.version.product.title} {result.plan.version.label}"
         logger.info(
-            f"event={job_type} context={context} status={log_status} duration={duration}"
+            "Job reached a completion status",
+            extra={
+                "context": {
+                    "event": f"{job_type}",
+                    "context": context,
+                    "status": f"{log_status}",
+                    "duration": duration,
+                }
+            },
         )
         result.save()
 
@@ -199,7 +207,7 @@ def run_flows(
     commit_ish = plan.commit_ish or plan.version.commit_ish
 
     with contextlib.ExitStack() as stack:
-        stack.enter_context(finalize_result(plan, result))
+        stack.enter_context(finalize_result(result))
         if result.user:
             stack.enter_context(report_errors_to(result.user))
         if scratch_org:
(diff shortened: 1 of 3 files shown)


metadeploy/api/jobs.py
```diff
@@ -49,12 +51,39 @@ def job(*args, **kw):
...
+    SUCCESS = "success"
+    FAILURE = "failure"  # preflight returned negative result
+    ERROR = "error"  # job threw an exception and failed
+    CANCELED = "canceled"  # by admins
+    TERMINATED = "terminated"  # by Heroku
+
+
 @contextlib.contextmanager
-def finalize_result(result):
+def finalize_result(plan: Plan, result: Union[Job, PreflightResult]):
```
davisagli, line 69: You can use `result.plan` rather than passing in the plan separately. And that will guarantee that we're using the plan that is related to the Job or PreflightResult rather than an arbitrary one.

metadeploy/api/jobs.py
```diff
@@ -74,17 +105,35 @@ def finalize_result(result):
...
+        # elif
+        if isinstance(result, PreflightResult):
+            job_type = JobType.PREFLIGHT
+        elif isinstance(result, Job):
+            job_type = JobType.JOB
+
+        context = f"{plan.version.product.title} {plan.version.label}"
+        logger.info(
+            f"event={job_type} context={context} status={log_status} duration={duration}"
+        )
```
davisagli, line 136: I think the context needs to be quoted if it contains spaces. We're trying to use logfmt style https://brandur.org/logfmt

davisagli: Actually I think metadeploy is already using a log formatter that will output logfmt if you pass extra values in the expected way: https://github.com/SFDO-Tooling/MetaDeploy/blob/main/metadeploy/logfmt.py#L22

metadeploy/api/tests/jobs.py
```diff
@@ -196,52 +197,98 @@ def test_finalize_result_worker_died(job_factory):
...
+
+@pytest.mark.django_db
+def test_finalize_result_job_success(job_factory, plan_factory, caplog):
+    job = job_factory(org_id="00Dxxxxxxxxxxxxxxx")
+    plan = plan_factory()
+    with finalize_result(plan, job):
+        pass
+
+    context = f"{plan.version.product.title} {plan.version.label}"
+    assert f"event=job context={context} status=success" in caplog.text
```
davisagli, line 289: It would be easier to verify the format is correct if you didn't use string formatting here. i.e. create test records with known values, and then assert that the output matches a literal string, rather than constructing the expected result from something else.

davisagli (changes requested):

davisagli (approved):

metadeploy/api/jobs.py
```diff
@@ -74,17 +105,43 @@ def finalize_result(result):
...
         if hasattr(e, "response"):
             result.exception += "\n" + e.response.text
         logger.error(f"{result._meta.model_name} {result.id} failed.")
         raise
     finally:
+        duration = (end_time - start_time).seconds
+        # TBD, dependent on another story.
+        # if result.is_release_test:
+        #    job_type = JobType.TEST_JOB
+        # elif
```
davisagli, line 127: @Br4nd0R please add this when you merge this into your work

PR #3226, opened 2021-10-08
Adjustments to job logging

davisagli:
- Make sure the event field is emitted at the beginning of the log line (needed in order to be turned into a metric by our log drain app)
- Provide a more human-friendly msg field
- Fixed a number of issues with quoting of logfmt field values
- Removed tag field, which we weren't using

### DIFF
--- a/metadeploy/logfmt.py
+++ b/metadeploy/logfmt.py
@@ -82,16 +82,10 @@ class LogfmtFormatter(ServerFormatter):
             or "unknown"
         )
 
-    def _get_tag(self, record):
-        tag = getattr(record, "tag", None)
-        if tag:
-            return quote_logvalue(tag)
-
     def format(self, record):
         parsed_msg = record.module == "logging_middleware"
         id_ = self._get_id(record)
         time = self._get_time(record)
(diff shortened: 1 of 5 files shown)


metadeploy/logfmt.py
```diff
@@ -85,19 +95,24 @@ def format(self, record):
...
         else:
-            msg = self._escape_quotes(record.getMessage())
-        rest = self.format_line(getattr(record, "context", {}))
-        fields = [
+            msg = record.getMessage()
+        context = getattr(record, "context", {})
+        rest = self.format_line(context)
+        fields = []
+        if "event" in context:
+            fields.append(f"event={context['event']}")
```
davidmreed, line 103: Any possibility `event` could include a space or quote in the future?

davidmreed (dismissed):

metadeploy/logfmt.py
davisagli: https://github.com/SFDO-Tooling/MetaDeploy/pull/3226/commits/dd91d1de6ce917b2daa52f5eafc13120d6b68451

PR #3227, opened 2021-10-08
Command to run plans against a scratch org

Br4nd0R:
# Changes
* There is a new command `run_plan` which takes a plan.id as an argument, and runs any preflight check and plan steps against a fresh scratch org.


metadeploy/api/jobs.py
```diff
@@ -331,9 +331,14 @@ def expire_preflights():
 expire_preflights_job = job(expire_preflights)
 
 
-def create_scratch_org(org_pk):
+def create_scratch_org(org_pk, release_test=False):
```
Br4nd0R, line 334: I know we talked about moving the release test logic out of this job. I would need to copy the first bit of login from this where the local github checkout is done into a separate function as well. Let me know what you think.

metadeploy/api/management/commands/run_plan.py
```diff
@@ -0,0 +1,31 @@
...
+            raise CommandError(f"Plan with Id {plan_id} does not exist.")
+
+        scratch_org = ScratchOrg.objects.create(
+            plan=plan, enqueued_at=datetime.utcnow().isoformat()
+        )
+        create_scratch_org(scratch_org.pk, release_test=True)
+
+        # If we delete the org here, then we are unable to
+        # query for it in the test
+        delete_scratch_org(scratch_org)
```
Br4nd0R, line 31: Should we make passing the `should_delete_locally` param conditional? I think we discussed that we may want to keep ScratchOrg models that fail around to be able to troubleshoot.

metadeploy/api/jobs.py
```diff
@@ -364,19 +369,22 @@ def create_scratch_org(org_pk):
                     duration=plan.scratch_org_duration,
                 )
         except Exception as e:
+            print(f"exception: {e}")
```
davisagli, line 372: left over debugging?

metadeploy/api/jobs.py
davisagli: Not a must-do, but I think that would help keep this function (focused on the customer scratch org scenario) tidier.

metadeploy/api/management/commands/run_plan.py
davisagli: +1 for keeping them around locally if the ScratchOrg has an error stored, because that's currently the only place it is recorded.

metadeploy/api/management/tests/commands.py
```diff
@@ -0,0 +1,60 @@
...
+                return_value=(org_config, None, None),
+            )
+        )
+        stack.enter_context(patch("metadeploy.api.jobs.delete_scratch_org_on_sf"))
+
+        call_command("run_plan", str(plan.id))
+
+    # We can't query the scratch org since it's been deleted
+    job = Job.objects.filter()[0]
+    preflight_result = PreflightResult.objects.filter()[0]
```
davisagli, line 47:
```suggestion
    job = Job.objects.get()
    preflight_result = PreflightResult.objects.get()
```

metadeploy/api/jobs.py
```diff
@@ -331,9 +331,14 @@ def expire_preflights():
 expire_preflights_job = job(expire_preflights)
```
davisagli, line 332: Don't forget to update lines 124-127 in this file (David Reed added them but commented out since the fields didn't exist yet)

davisagli (approved):

PR #3268, opened 2021-10-18
Week 41 Updates

TheBitShepherd:


requirements/dev.txt
```diff
@@ -35,7 +35,7 @@ click==7.1.2
     #   -c requirements/prod.txt
     #   black
     #   pip-tools
-coverage==5.5
+coverage[toml]==5.5
```
codexmarcel, line 38: Let's watch this.

Br4nd0R: Good catch! This is a transient dev dependency, so _will _not__ impact production environments. I have a suspicion that this is something automated `pip-tools` is doing.

PR #3338, opened 2021-11-08
Make sure unhandled exceptions in preflight tasks get logged

davisagli:
Also make sure that unhandled exceptions from both Jobs and Preflights store a traceback in the exception field.

As an example, before this change, we merely got an unhelpful line in the `log` field: Exception in task get_installed_packages

Now we get a full traceback and error message in the `exception` field:
Screen Shot 2021-11-08 at 11 17 21 AM

### DIFF
--- a/metadeploy/api/flows.py
+++ b/metadeploy/api/flows.py
@@ -134,13 +134,12 @@ class PreflightFlowCallback(BasicFlowCallback):
         self.context.save()
 
     def post_task(self, step, result):
-        """Report exception evaluating a preflight task."""
+        """Report exception evaluating a preflight task.
+
+        If there was an exception while running the task,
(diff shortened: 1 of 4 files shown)


metadeploy/api/flows.py
```diff
@@ -134,13 +134,12 @@ def post_flow(self, coordinator):
...
+        """
         if result.exception:
-            error_result = {
-                "plan": {
-                    "status": ERROR,
-                    "message": bleach.clean(str(result.exception)),
-                }
-            }
-            self.context.results.update(error_result)
-            self.context.save()
```
davisagli, line 146: This removed part was trying to show the exception message to the customer, but it wasn't actually working because it gets overwritten in `post_flow`

Br4nd0R (approved):

PR #3359, opened 2021-11-10
Week 45 Updates

Br4nd0R:
* No formatting updates from `black` to commit.


davisagli (commented):
need to wait for the new cumulusci

davisagli (approved):

PR #3368, opened 2021-11-15
🛢 Auto-publish installers at publish datetime

bethbrains:


metadeploy/api/jobs.py
```diff
@@ -296,6 +296,16 @@ def enqueuer():
 enqueuer_job = job(enqueuer)
 
 
+def publish_installers():
+    now = timezone.now()
+    for v in Version.objects.filter(publish_date__lt=now, is_listed=False):
```
davisagli, line 301: Does this include publish_date of None or not? We don't want to.

metadeploy/api/serializers.py
```diff
@@ -271,6 +271,7 @@ class Meta:
             "primary_plan",
             "secondary_plan",
             "is_listed",
+            "publish_date",
```
davisagli, line 274: @davidmreed Does the frontend need access to publish_date? Admin access is via a separate API in https://github.com/SFDO-Tooling/MetaDeploy/blob/main/metadeploy/adminapi/api.py and should automatically include the new field.

davisagli:
(Added do-not-merge label to remind us this is pending security assessment)

metadeploy/api/jobs.py
bethbrains: Added a test case to cover this.

metadeploy/api/serializers.py
davidmreed: No, it doesn't. Thanks for clarifying that! I did not realize the admin API was separate.

davidmreed (approved):

PR #3369, opened 2021-11-16
limit plans to begin regression testing with

Br4nd0R:


metadeploy/api/management/tests/commands.py
```diff
@@ -100,7 +100,7 @@ def test_get_plans_to_test(plan_template_factory, plan_factory):
     plan_factory(tier=Plan.Tier.primary, plan_template=template_opted_out)
 
     plans_to_test = get_plans_to_test()
-    assert len(plans_to_test) == 2
+    assert len(plans_to_test) == 1
```
Br4nd0R, line 103: @davisagli - Are you aware of a way to recreate getting None values for plans in the DB?

davisagli: @Br4nd0R I think you already are getting None for template_2 since it doesn't have primary plans

davisagli (approved):

PR #3386, opened 2021-11-17
disable non-security dependabot PRs

davisagli:


Br4nd0R (approved):
This one's a good one. 😎

PR #3389, opened 2021-12-02
Log error and context on scratch org creation failure

Br4nd0R:
We now log an error message along with additional context if a failure during scratch org creation occurs.


Br4nd0R:
Would if be worth including the error message from the exception in the context somewhere?

davisagli (approved):
LGTM

PR #3396, opened 2022-01-14
Updates for Running on Docker

Br4nd0R:
# Changes
* Re-formatted `.env.docker.example` with headers for a cleaner layout
* `.env.docker.example` now includes the `DEVHUB_USERNAME` variable.
* `.env.docker.example` now includes the `GITHUB_TOKEN` variable.
* Fixed an issue where a raw newline char (`\n`) was not displaying properly.
* Added `DEVHUB_USERNAME` to `running_docker.rst` noting how to find the proper value.


Br4nd0R:
I see `GITHUB_APP_ID` and `GITHUB_APP_KEY` in `.env.docker.example` but no mention of them in `running_docker.rst`. Are these only needed for a production environment where you would like access to the GitHub API via a GitHub app, or something else?

I do see these variables mentioned in [running.rst](https://github.com/SFDO-Tooling/MetaDeploy/blob/main/docs/running.rst), but no specifics on what they're used for.

.env.docker.example
```diff
@@ -1,27 +1,42 @@
...
+#########################
+# Ensure the following three match with docker-compose.yml
 POSTGRES_DB=metadeploy
 POSTGRES_USER=postgres
 POSTGRES_PASSWORD=
 
-# General settings
+# Change this to a new key,
+# generated using cryptography.fernet.Fernet.generate_key()
+DB_ENCRYPTION_KEY=Ul-OySkEawSxUc7Ck13Twu2109IzIFh54C1WXO9KAFE=
```
davisagli, line 11: We must never include secrets in the repository, even if they are not intended for production use.

.env.docker.example
```diff
@@ -1,27 +1,42 @@
...
 DJANGO_DEBUG=True
 
 SECURE_SSL_REDIRECT=False
 
-# Application Settings
+#########################
+#  Salesforce Settings
+#########################
+DEVHUB_USERNAME=...
+DOCKER_SFDX_HUB_KEY=...
```
davisagli, line 29: We should include a comment that these 2 are only needed if you want to enable the scratch org feature (and comment them out by default?)

.env.docker.example
Br4nd0R: I was under the impression that this was an acceptable sample since it has been in repo for 2 years. I can go ahead and remove this.

davisagli: @Br4nd0R I know you didn't change it, but moving it made me notice that it's not following best practice.

davisagli (dismissed):

davisagli (approved):

PR #3398, opened 2022-01-18
Fixing Async error with Scratch Org failures

davidjray:
Fixing [W-10387614](https://gus.lightning.force.com/a07EE00000cnH5BYAU)


metadeploy/api/tests/push.py
```diff
@@ -53,14 +54,33 @@ async def test_notify_org_result_changed(
...
+    from ..serializers import ScratchOrgSerializer
+
+    scratch_org_factory = sync_to_async(scratch_org_factory)
+    soj = await scratch_org_factory()
+
+    notify_org_mock = mocker.patch("metadeploy.api.push.notify_org")
+
+    await notify_org_changed(soj)
+
+    expected_payload = ScratchOrgSerializer(soj).data
```
Br4nd0R, line 77: @davisagli - @davidjray and I have verified that your original hypothesis for the bug fix works in our local environments. We've stashed those changes and are attempting to write a test that now _fails_ with the error that was encountered to be sure the test is legitimate.

We would expect that since this test is running in an async context, that it should fail on this line. It does not. Likewise, if we make the assert below simpler (`assert_called_once()`) and remove this line, the test passes. We have verified that it does in-fact reach the point of serializing the scratch org in the [production code base](https://github.com/SFDO-Tooling/MetaDeploy/blob/e930f549d56135d9144de5a7b684cc493f150191/metadeploy/api/push.py#L224). This is the point where we would expect to see the original `SynchronousOnlyOperation` exception but we don't. Any guidance would be much appreciated.

davisagli: @Br4nd0R @davidjray hmm. It sounds like in the production scenario the serializer is triggering a database query, but in the test scenario it is not. That's likely due to some difference with the scratch org object constructed by django-factoryboy, but I'm not sure if it's because it's in a different state, or because it has prefetched some data that needs to be fetched on demand in the production scenario. Does the stack trace give any insight into what in particular the serializer is trying to query?

davisagli: @Br4nd0R @davidjray I can't tell from the stack trace, but my guess is that it's because of the ScratchOrg's `plan` -- my theory is that in production the plan needs to be fetched from the db, but in the test it has already been loaded when the factory constructed it. If that hunch is correct, you could probably recreate the production scenario by changing how you get the`soj` object here to do a fresh query rather than using the return value from the factory.

davisagli (dismissed):

davisagli (approved):

PR #3399, opened 2022-01-19
Week 3 updates

Br4nd0R:


davisagli:
@Br4nd0R looks like this has the lint errors that happen when there are 2 copies of `@types/react`

Br4nd0R:
Not sure what's up with the linting error here...

davisagli (approved):

PR #3409, opened 2022-03-11
Week 10 Updates

Br4nd0R:
There are a handful of JS packages that are in the red (have major version updates to complete) and some of these cause tests to break. I went ahead and compared these to the current versions in Metecho, and all but one were at the same versions. I went ahead an updated the single critical update that was out-of-sync with Metecho (`eslint-plugin-jest` to version `26.1.1`).


jstvz (approved):
LGTM

PR #3423, opened 2022-06-09
Fix bug when re-running expired preflight checks

Br4nd0R:
* Fixed a bug where re-running expired preflight checks would hang indefinitely.


davidmreed (approved):
Nice find and nice fix. 

Did you create a WI to use the field subclass approach to permanently deal with this, or shall I?

Br4nd0R:
@davidmreed - Already created: W-11268662

PR #3424, opened 2022-06-09
Support all ScratchOrgInfo keys when generating a scratch org

davidmreed:
MetaDeploy now supports all keys on the ScratchOrgInfo object when generating a scratch org.

Note that objectSettings is not currently supported because of a weakness in the underlying CumulusCI `OrgSettings` task. A separate PR will cure that with no further changes to MetaDeploy.

### DIFF
--- a/metadeploy/api/salesforce.py
+++ b/metadeploy/api/salesforce.py
@@ -16,6 +16,7 @@ from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import gettext_lazy as _
 from requests.exceptions import HTTPError
+from requests.structures import CaseInsensitiveDict
 from rq import get_current_job
 from simple_salesforce import Salesforce as SimpleSalesforce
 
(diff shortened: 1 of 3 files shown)


davidmreed:
Pending E2E test

metadeploy/api/salesforce.py
```diff
@@ -149,27 +150,45 @@ def _get_org_result(
...
+    for field in fields:
+        if field in scratch_org_definition:
+            create_args[field] = scratch_org_definition[field]
+
+    # Lastly, populate default items.
+    # Note that requests' CaseInsensitiveDict does not support setdefault()
+    if "OrgName" not in create_args:
+        create_args["OrgName"] = "MetaDeploy Scratch Org"
+    if "HasSampleData" not in create_args:
+        create_args["HasSampleData"] = False
```
jstvz, line 191: ❓: Why wouldn't replacing this with `set.difference()` and `dict.update()` work here?

davidmreed: I rewrote this section to use a different strategy.

davidmreed:
E2E tests pass on local. Also fixes a previously unknown bug: failing to poll for scratch org completion that takes more than an unspecified very short time.

metadeploy/api/salesforce.py
```diff
@@ -149,35 +150,63 @@ def _get_org_result(
...
     if SFDX_SIGNUP_INSTANCE:  # pragma: nocover
-        create_args["Instance"] = SFDX_SIGNUP_INSTANCE
+        create_args["instance"] = SFDX_SIGNUP_INSTANCE
     response = devhub_api.ScratchOrgInfo.create(create_args)
 
     # Get details and update scratch org config
     return devhub_api.ScratchOrgInfo.get(response["id"])
 
 
+def _poll_for_scratch_org_completion(devhub_api, org_result):
```
Br4nd0R, line 199: It looks like this logic wasn't present previously. So this fixed a bug where we might have issued a `GET` for a record before it was finished being created?

Br4nd0R (approved):
LGTM!

metadeploy/api/salesforce.py
davidmreed: Exactly. We were getting back `None` for `AuthCode` because the org creation wasn't actually done.

jstvz (commented):
Crud, forgot to hit send on this, sorry.

metadeploy/api/salesforce.py
```diff
@@ -149,35 +150,63 @@ def _get_org_result(
...
+    while org_result["Status"] in ["New", "Creating"]:
+        time.sleep(10)
+        org_result = devhub_api.ScratchOrgInfo.get(org_result["Id"])
```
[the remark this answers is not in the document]
davidmreed: Yeah, I can't assume the platform will behave itself, can I. Will revise, and also check the timeout on the queue where this is running since big orgs can take > 1 hour to clone.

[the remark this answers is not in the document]
davidmreed: After reviewing, now want to take back that suggestion. Because MetaDeploy has a `default`-queue timeout of 3600 seconds the job will simply be terminated if the org create doesn't finish in a reasonable time.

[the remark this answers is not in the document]
davidmreed: I'll just use reuse the default job timeout value here in case we start running builds in one-off dynos.

PR #3427, opened 2022-06-09
Update PreflightResult queryset

Br4nd0R:


metadeploy/api/models.py
```diff
@@ -862,7 +862,7 @@ def run(self, ctx, plan, steps, org):
 
 class PreflightResultQuerySet(models.QuerySet):
     def most_recent(self, *, org_id, plan, is_valid_and_complete=True):
-        kwargs = {"org_id": org_id, "plan": plan}
+        kwargs = {"org_id__startswith": org_id, "plan": plan}
```
davidmreed, line 865: Is the org id coming directly from the frontend here, or is it being validated by us before this method call? If it's under user control, switching to `startswith` has a potential for information leak.

PR #3428, opened 2022-06-09
Assistive Text update for progress indicator

davidjray:
Move the SLDS assistive text to the element designated as the label. This acts as a dynamic progress indicator, and gives users a better experience.

### DIFF
--- a/src/js/components/jobs/progressBar.tsx
+++ b/src/js/components/jobs/progressBar.tsx
@@ -127,6 +127,9 @@ class ProgressBar extends React.Component<Props, State> {
           <span>
             <strong>
               {t('Installation Progress')}
+              <span className="slds-assistive-text">
+                {t('{{percent}}% Complete', { percent: progressRounded })}
+              </span>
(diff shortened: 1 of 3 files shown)


src/js/components/jobs/progressBar.tsx
```diff
@@ -127,6 +127,9 @@ class ProgressBar extends React.Component<Props, State> {
           <span>
             <strong>
               {t('Installation Progress')}
+              <span className="slds-assistive-text">
+                {t('{{percent}}% Complete', { percent: progressRounded })}
```
bethbrains, line 131: does this need to be added to the translation file?

bethbrains (changes requested):

src/js/components/jobs/progressBar.tsx
davidjray: @bethbrains nope! it's already included 👍🏽

bethbrains (approved):

PR #3456, opened 2022-08-08
Dependency rollback

jstvz:
We still need to bisect to find the culprit, but we have to unblock a customer in the meantime.

Commits:

- Roll all dependencies back
- Upgrade CumulusCI to 3.63.0
- Add Vlocity as JS dependency


davidjray (approved):

requirements/prod.txt
```diff
@@ -233,13 +233,13 @@ markupsafe==2.0.1
     #   snowfakery
 msgpack==1.0.4
     # via channels-redis
-newrelic==7.14.0.177
```
jgerigmeyer, line 236: I'm not entirely sure what client bug you're trying to solve here, but FWIW I just discovered that this specific NewRelic update broke websocket events in unexpected ways. Pinning this back (or removing newrelic entirely?) fixes a number of related issues.

jstvz: Thanks for the report. We missed your post and had to figure it out the hard way.

jstvz:
Closing in favor of #3458 and #3459

requirements/prod.txt
jgerigmeyer: Sorry! Sounds like I wouldn't have saved you much if any time in the end.

PR #3457, opened 2022-08-25
Fix `expire_preflights` Job

Br4nd0R:
* Fixed an issue where stale preflights in a `is_valid` state with a status other than `completed` were preventing customer's from clicking the "Install" button on any product. The `expire_preflights` job will now expire _all_ preflights that are older than `PREFLIGHT_LIFETIME_MINUTES` regardless of status.


jstvz (dismissed):

jstvz (approved):
Squash or be forever haunted by the ghosts of your intermediate commits. 😏

jstvz:
👻

PR #3459, opened 2022-08-26
Add OmniScript dependencies

jstvz:
Required to support [local compilation] of FlexCards in Vlocity CLI. This requires the addition of a private NPM repository ([Heroku docs])to allow the installation of the compiler via:

```console
yarn add @omniscript/omniscript-lwc-compiler
```

There are alternative routes specified in the Vlocity documentation, but users reported they were unsuccessful.

[Heroku docs]: https://devcenter.heroku.com/articles/nodejs-support#private-dependencies
[local compilation]: https://github.com/vlocityinc/vlocity_build#initial-support-for-omniscript--flexcards-local-compilation

### DIFF
--- a/.github/setup-docker-compose/action.yml
+++ b/.github/setup-docker-compose/action.yml
@@ -35,6 +35,8 @@ runs:
     - uses: docker/setup-buildx-action@v1
     - name: Build
       uses: docker/build-push-action@v2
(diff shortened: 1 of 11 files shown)


jstvz:
Using `puppeteer` in #3468 instead.

PR #3460, opened 2022-08-29
Clean Up Cruft

Br4nd0R:
* Removed `.pyup.yml`
* Remove `locale/` directory
* Moved `.browserslistrc` into `package.json`

I was not able to verify if translations were working locally after removing the `locale/` dir (although it is not referenced in the codebase). Will need to verify on staging.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,7 +1,0 @@
-> 0.5%
-last 2 versions
-Firefox ESR
-not dead
-not ie 11
-not ie_mob 11
-not op_mini all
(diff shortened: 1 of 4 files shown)


jgerigmeyer:
> I was not able to verify if translations were working locally after removing the `locale/` dir (although it is not referenced in the codebase). Will need to verify on staging.

@Br4nd0R Front-end (JS/React) translations will continue to work fine without the `locale/` dir -- that's intended for translated back-end (Python) user-facing strings. The translation files can be generated with `python manage.py makemessages --locale <locale>`, and then compiled with `python manage.py compilemessages`. (I don't think those have been generated or translated ever, AFAIK.) See: https://github.com/SFDO-Tooling/MetaDeploy/blob/main/CONTRIBUTING.md#internationalization

PR #3466, opened 2022-09-30
Protect certain users from auto deletion

davidjray:


metadeploy/api/migrations/0123_user_do_not_delete.py
```diff
@@ -0,0 +1,21 @@
...
+        ("api", "0122_add_multi_tenancy"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="do_not_delete",
+            field=models.BooleanField(
+                default=False,
+                help_text="Do not delete User during monthly User clean up job",
```
TheBitShepherd, line 18:
```suggestion
                help_text="Denotes whether a user should be skipped when the delete_old_users job runs.",
```

metadeploy/api/models.py
```diff
@@ -162,6 +162,11 @@ class UserManager(BaseUserManager):
 class User(HashIdMixin, AbstractUser):
     objects = UserManager()
 
+    do_not_delete = models.BooleanField(
+        default=False,
+        help_text="Do not delete User during monthly User clean up job",
```
TheBitShepherd, line 167:
```suggestion
        help_text="Denotes whether a user should be skipped when the delete_old_users job executes.",
```

metadeploy/api/tests/test_cleanup.py
```diff
@@ -49,12 +49,14 @@ def test_delete_old_users(user_factory):
...
     old_user = user_factory(last_login=two_months_ago)
     staff_user = user_factory(last_login=two_months_ago, is_staff=True)
+    another_user = user_factory(last_login=two_months_ago, do_not_delete=True)
 
     delete_old_users()
 
     # make sure only the old user was deleted
     new_user.refresh_from_db()
     staff_user.refresh_from_db()
+    another_user.refresh_from_db()
```
TheBitShepherd, line 59: 👍

metadeploy/api/models.py
```diff
@@ -162,6 +162,11 @@ class UserManager(BaseUserManager):
 class User(HashIdMixin, AbstractUser):
     objects = UserManager()
 
+    do_not_delete = models.BooleanField(
```
TheBitShepherd, line 165: I'm *slightly* against using negatives in the name of a boolean field. I think something like `is_persistent` is slightly easier to read when parsing through the code. If you want to make this change feel, but no worries if not.

metadeploy/api/admin.py
```diff
@@ -284,7 +284,51 @@ class UserAdmin(AdminHelpTextMixin, BaseUserAdmin):
         "GDPR reminder: The username, name, and email are personally identifiable information. "
         "They must be used for support/debugging purposes only, and not exported from this system."
     )
-    list_display = ("username", "is_active", "is_staff", "is_superuser", "date_joined")
+    fieldsets = (
```
TheBitShepherd, line 287: Were we not able to just add the new field to the `list_display`? What do these changes do? Looks like these might have been some "nice to have" additions while you were poking around in the admin UI?

TheBitShepherd (changes requested):

metadeploy/api/tests/test_cleanup.py
TheBitShepherd: TIL: `refresh_from_db()` throws an error if the record has been deleted.

metadeploy/api/admin.py
davidjray: I could not get the new field to display without doing it this way. As a bonus of this method, the field also exists on the Create New User form as well, so it adds some improvement.

Similar situation in [this](https://stackoverflow.com/questions/66102893/custom-field-not-showing-in-admin-page) thread.

metadeploy/api/models.py
```diff
@@ -162,6 +162,11 @@ class UserManager(BaseUserManager):
 class User(HashIdMixin, AbstractUser):
     objects = UserManager()
 
+    is_persistent = models.BooleanField(
```
TheBitShepherd, line 165: 🎉

TheBitShepherd (approved):

PR #3468, opened 2022-10-05
Use Puppeteer VBT compiler in workers

prescod:
Make a [new script](https://github.com/SFDO-Tooling/MetaDeploy/pull/3468/files#diff-568838017c33edbf49458e93c994c192f6c774bae92d345172ad082e3b144983R1) that makes Chrome available to the VBT compiler.

Invoke it from [Procfile](https://github.com/SFDO-Tooling/MetaDeploy/pull/3468/files#diff-0a99231995da379e7aebabe76c9d849a23737a42c3b3a8994043e2aa80958424R3). 

Install Vlocity and Upgrade "[MiniPass](https://github.com/SFDO-Tooling/MetaDeploy/pull/3468/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)" to work around a bug that seems to have arisen asynchronously (perhaps the type declarations are downloaded at runtime).

### DIFF
--- a/.heroku/start_metadeploy_worker.sh
+++ b/.heroku/start_metadeploy_worker.sh
@@ -1,0 +1,4 @@
+set -e
+mkdir -p /opt/google/chrome
+ln -s /app/.apt/usr/bin/google-chrome /opt/google/chrome/chrome
+python manage.py rqworker default
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "i18next-http-backend": "^1.4.1",
     "js-cookie": "^3.0.1",
     "lodash": "^4.17.21",
+    "minipass": "^3.3.5",
     "react": "^18.2.0",
     "react-document-title": "^2.0.3",
     "react-dom": "^18.2.0",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "i18next-http-backend": "^1.4.1",
     "js-cookie": "^3.0.1",
     "lodash": "^4.17.21",
+    "minipass": "^3.3.5",
```
jstvz, line 66: Did you try setting `resolutions` ([docs](https://classic.yarnpkg.com/en/docs/selective-version-resolutions))?

.heroku/start_metadeploy_worker.sh
```diff
@@ -0,0 +1,4 @@
+set -e
+mkdir -p /opt/google/chrome
+ln -s /app/.apt/usr/bin/google-chrome /opt/google/chrome/chrome
```
jstvz, line 3: 😑

jstvz (approved):

package.json
prescod: Nope. I didn't know about that. Too bad pip doesn't have it!

jstvz (approved):

PR #3469, opened 2022-10-05
Update token endpoint to enforce tenanted tokens

davidjray:
Rework OptainTokenView to be RefreshTokenView to not allow users to obtain tokens for a tenant they are not already authorized to. The endpoint will allow users to reset an existing token when authorized via a username and password.

### DIFF
--- a/metadeploy/api/tests/test_views.py
+++ b/metadeploy/api/tests/test_views.py
@@ -68,20 +68,36 @@ class TestUserView:
 
 
 @pytest.mark.django_db
-class TestObtainTokenView:
+class TestResetTokenView:
+    def test_reset(self, anon_client, user_factory, extra_site):
+        url = reverse("token")
+        user = user_factory()
+        data = {"username": user.username, "password": "foobar"}
+
+        token1 = Token.objects.create(user=user)
+
+        response = anon_client.post(url, data=data)
+        assert response.data.get("token") is not token1.key
+
+    def test_not_authorized(self, anon_client, user_factory, extra_site):
+        url = reverse("token")
+        user = user_factory()
+        data = {"username": user.username, "password": "foobar"}
+
+        response = anon_client.post(url, data=data)
+        assert response.status_code == 403
+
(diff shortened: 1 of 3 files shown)


metadeploy/api/tests/test_views.py
```diff
@@ -68,20 +68,39 @@ def test_multi_tenancy(self, client, extra_site):
...
+        old_token = Token.objects.create(user=user)
+        response = client.get(url)
+
+        tokens = Token.objects.filter(user=user)
+        # Refreshing should not add another token
+        assert tokens.count() == 1
+
+        new_token = tokens[0]
+        assert response.data.get("token") == new_token.key
+        assert old_token.key != new_token.key
```
TheBitShepherd, line 84: Let's assert that the token's are completely different instances:
```suggestion
        assert new_token is not old_token
```

metadeploy/api/tests/test_views.py
```diff
@@ -68,20 +68,39 @@ def test_multi_tenancy(self, client, extra_site):
...
+        assert response.status_code == 403
+        assert str(response.data) == "Unable to reset token: a token does not exist."
 
-        response = anon_client.post(url, data=data, SERVER_NAME=extra_site.domain)
-        assert response.data.get("token") == token2.key
+    def test_multi_tenancy(self, client, user_factory, extra_site):
+        url = reverse("token")
+
+        # Generate token for user under default tenant
+        Token.objects.create(user=client.user)
```
TheBitShepherd, line 98: Let's add a `get` call after this to the default tenant, and assert that we get a 200 back. This way we are showing that it works for the tenant the user has a token for, and throws a 403 for the one we don't have a token for.

You could also just add lines 101-103 below at the bottom of the `test_reset_token` method, and it would do essentially the same thing.

TheBitShepherd (changes requested):

metadeploy/api/tests/test_views.py
davidjray: Done.

TheBitShepherd (approved):

PR #3482, opened 2022-12-13
added an aria-label for the link surrounding a product card

boakley:
added an aria-label for the link surrounding a product card
See #W-11149827

A couple of things I'm not sure of:
* I set the aria-label on the Link component. Other choices were the outer div or the inner Card. 
* I set the label text to be "Product Link: <product name> version <version>". Is that too wordy?
* I only added a translation for "Product Link:" rather than the entire label. Is that correct or should the entire string be translated?

### DIFF
--- a/src/js/components/products/listItem.tsx
+++ b/src/js/components/products/listItem.tsx
@@ -24,7 +24,9 @@ const ProductItem = ({ item }: { item: Product }) => {
       <Link
         to={routes.product_detail(item.slug)}
         className="slds-text-link_reset"
-        aria-label={`${item.title} ${t('Version {{version}}', { version: label })}`}
+        aria-label={`${item.title} ${t('Version {{version}}', {
+          version: label,
+        })}`}
       >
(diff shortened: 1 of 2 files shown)


src/js/components/products/listItem.tsx
```diff
@@ -24,6 +24,7 @@ const ProductItem = ({ item }: { item: Product }) => {
       <Link
         to={routes.product_detail(item.slug)}
         className="slds-text-link_reset"
+        aria-label={`${t('Product Link')}: ${item.title} version ${label}`}
```
davidjray, line 27: In my testing this works great. 

Screenshot 2022-12-14 at 9 22 22 AM

My only suggested change is to remove the "Product Link" text. I think that is a good idea, but most likely it is not needed, since the screenreader already calls it a link.

davidjray (commented):
The label works great, just one small change requested.

davidjray (approved):

PR #3490, opened 2023-01-13
front-end bootstrap endpoint

TheBitShepherd:
New bootstrap endpoint for the front-end is available at: `/api/ui/boostrap`


TheBitShepherd:
@davidjray - Let me know if you would also like the test to assert on the presence of particular values. I believe the work item  specified that if `SiteProfile.company_name` was null, that it would be replaced with `"MetaDeploy"`. Do we still want this behavior?

davidjray (approved):

PR #3501, opened 2023-03-03
Make Contents of `django-js-reverse` available via API.

TheBitShepherd:
* MetaDeploy now serves the contents of `staticfiles/django_js_reverse/js/reverse.js` under the `/api/ui/reverse` endpoint.

### DIFF
--- a/metadeploy/api/tests/test_views.py
+++ b/metadeploy/api/tests/test_views.py
@@ -1,5 +1,6 @@
 from contextlib import ExitStack
 from datetime import datetime
+from pathlib import Path
 from unittest.mock import MagicMock, patch
 from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


davidjray (dismissed):

metadeploy/api/tests/test_views.py
```diff
@@ -908,3 +909,21 @@ def test_front_end_info(anon_client, site_profile_factory):
 
     actual_site_keys = content["SITE"].keys()
     assert expected_site_keys == set(actual_site_keys)
+
+
+@pytest.mark.django_db
+def test_reverse_endpoint(anon_client):
```
jstvz, line 915: ↔️ : check out the `tmp_path`  [fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture). Main advantage is that it'll tear it down for you. You can also pass `encoding=encoding` to [`Path.write_text`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text).

jstvz (dismissed):

metadeploy/api/tests/test_views.py
TheBitShepherd: `tmp_path`.... that's what I was looking for!

davidjray (approved):

PR #3503, opened 2023-03-14
Set up basic back-end mocking

davidjray:
Leaning on mock-fetch, that we are already using, I setup a basic mocking framework that looks lin the data.js file, and if a mathci key is found it will mock the call. You can turn on the mocking by setting changing line 7 of `mock/mocking_util.js` for now. After launch we will work on a more permanent way to activate the mocking.


davidjray:
Closing this PR in favor of some enhancements to come later.

PR #3502, opened 2023-03-14
Initial setup of HTML file and bootstrap loading

davidjray:
Switched from using "react-document-title" to "react-helmet-async". React Helmet Async allow updates of the HTML template meta tags. So after loading the bootstrap endpoint and grabbing the globals we can update the HTML page meta tags dynamically, including the page title.

PR #3506, opened 2023-03-15
Move the processing of the user logged in/out status to api call

davidjray:
Merge AFTER #3502 

Remove the DOM injected User handling that was depended on Django.


jofsky (approved):
LGTM.

PR #3516, opened 2023-06-14
Remove mock DB_ENCRYPTION_KEY

jstvz:
Having the mock DB_ENCRYPTION_KEY in env.example resulted in false
positives during static analysis. This commit removes the mock key, and
improves documentation by including a working one-liner to generate a
valid fernet key.


jstvz:
@jofsky

jofsky (approved):

PR #3526, opened 2023-08-21
Annotate Dockerfile to avoid implication of a key disclosure

davidmreed:


jstvz (approved):
😑

jstvz:
Less facetiously: we can update [triage.yml](https://github.com/SFDO-Tooling/MetaDeploy/blob/main/sfdc_metadata/cred_scan_triage/triage.yaml), but only after there's a false positive from the credential scanner..

PR #3538, opened 2023-10-12
Bug Fixed for DRF showing in api documentation

jain-naman-sf:


davidjray (approved):
To be clear, this clears out the header, but does not replace it with anything. I am fine with that.

PR #3543, opened 2023-11-17
Add tenant IP allow list

pratibhavermasalesforce:
changes done to add new attribute allowed_ip_addresses to siteprofile and validating client ip against it


jkasturi-sf (approved):
This will allow admins to specify list of IP addresses from which Metadeploy can be accessible. This restriction will NOT be applied to Admin pages.

metadeploy/multitenancy/iprestrict_middleware.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        client_ip = request.META.get('REMOTE_ADDR', None)
+        profile = self.getSiteProfile()
+        is_admin_url = request.path.startswith(ADMIN_URL)
+
+        if not is_admin_url and hasattr(profile, "allowed_ip_addresses") and profile.allowed_ip_addresses:
+            if client_ip not in profile.allowed_ip_addresses:
```
jstvz, line 23: ↔ : My personal preference is to simplify conditionals wherever possible by extracting variables and/or combining `if` statements. Another approach could be:
```python
    def __call__(self, request):
        client_ip = request.META.get("REMOTE_ADDR", None)
        profile = self.getSiteProfile()
        is_admin_url = request.path.startswith(ADMIN_URL)
        has_ip_allowlist = (
            hasattr(profile, "allowed_ip_addresses") and profile.allowed_ip_addresses
        )
        should_filter = is_admin_url and has_ip_allowlist
        
        if should_filter and client_ip not in profile.allowed_ip_addresses:
            return HttpResponseForbidden(
                "You don't have permission to access this resource."
            )

        return self.get_response(request)
```

This adds more lines of code, but I think it improves readability.

jstvz (commented):
I see this PR is currently checking IP addresses directly, which means we can only allow or block specific IPs one by one. This method is simple (which is good) but can be limiting, especially if we need to handle a large range or changing IPs. 

Have you considered using CIDR notation for IP ranges? CIDR allows us to cover more IPs in a flexible way. Python’s [`ipaddress`](https://docs.python.org/3.9/library/ipaddress.html) module can help us implement this. It's a common approach and more scalable.

---
<details>
<summary>Emoji Key</summary>


| Emoji | Meaning                                          | Blocking |
|-------|--------------------------------------------------|----------|
| ±     | Requested change                                 | Yes      |
| ❓    | Query (Request for context or explanation)       | Yes      |
| ❔    | Query (Request for context or explanation)       | No       |
| 💭    | Note/Question to myself or the @-mentioned user. | No       |
| ↔     | "Here's another way to do this:"                 | No       |
| ⬇     | Aesthetic or stylistic quibble                   | No         |

</details>

metadeploy/multitenancy/iprestrict_middleware.py
pratibhavermasalesforce: I see your point, I agree, it does look more clean and readable. I'll make the change.

pratibhavermasalesforce:
Hey @jstvz , Regarding https://github.com/SFDO-Tooling/MetaDeploy/pull/3543#pullrequestreview-1782934935, I had a similar question and confirmed with @davidmreed , what I understood from him is as of now we want to keep it simple, maybe later we can enhance the functionality.

metadeploy/multitenancy/iprestrict_middleware.py
pratibhavermasalesforce: Done.

jkasturi-sf (approved):
It supports IP address alone and CIDR format to check the allow list

PR #3547, opened 2023-12-13
@W-14777545 - Update python deps

vsbharath:


requirements/prod.txt
```diff
@@ -4,102 +4,81 @@
...
     #   pyjwt
-    #   pyopenssl
     #   secretstorage
     #   service-identity
     #   sfdo-template-helpers
-cumulusci==3.76.0
+cumulusci==3.83.0
     # via -r requirements/prod.in
-daphne==3.0.2
-    # via channels
```
jstvz, line 95: [This](https://channels.readthedocs.io/en/stable/releases/4.0.0.html#decoupling-of-the-daphne-application-server) is the cause of your test failure. You can update `prod.in` to include `daphne` as an optional dependency:
```diff
-channels
+channels[daphne]
```

That should also pull in `twisted`.

requirements/prod.txt
```diff
@@ -4,102 +4,81 @@
...
 defusedxml==0.7.1
     # via
     #   cumulusci
     #   python3-openid
-deprecated==1.2.13
-    # via redis
-django==3.2.14
+django==4.2.8
```
jstvz, line 81: 👍🏾
Good to use the LTS, but let's pin it to a compatible release  in `prod.in`.

runtime.txt
```diff
@@ -1 +1 @@
-python-3.9.13
+python-3.10.13
```
jstvz, line 1: While I do want to move to 3.10 (3.12, really) ASAP, let's update that separately since it's a bit more involved (CI 
 and containers need to be updated, we need to run `pyupgrade`, etc.). We do need the [latest supported 3.9 version](https://devcenter.heroku.com/articles/python-support#supported-runtimes):
```suggestion
python-3.9.18
```

jstvz (changes requested):
Thanks! Needs some tweaks.

jstvz (approved):

PR #3557, opened 2024-03-19
Update user timeout warning @W-11149815

jstvz:
Applies the suggested remediation from the PAID team to improve
accessibility.


yippie:
@jstvz This commit seems to have broken the timeout warning. The warning now looks like this:

`If no action is performed for minutes, you will be automatically logged out and will need to log in again. minutes, you will be automatically logged out and will need to log in again.`

Notice the minutes is gone and the part of the message after the merge field is duplicated.

PR #3558, opened 2024-04-04
W14 dependency updates

jstvz:


yippie:
Please merge, the current version of Metadeploy is using cumulusci 3.84.3 which has some bugs in it we are hitting.

requirements/prod.txt
```diff
@@ -91,15 +90,15 @@ cryptography==41.0.7
     #   secretstorage
     #   service-identity
     #   sfdo-template-helpers
-cumulusci==3.84.3
+cumulusci==3.86.0
```
vsbharath, line 93: @jstvz This needs to be updated to the latest version which was released. We can update with latest dependencies

PR #3569, opened 2024-07-25
Validate websocket origin

jstvz:
WebSockets can be initiated from any site on the internet and still have the user’s cookies and session. This pull request restricts the sites which are allowed to open sockets to the app.

See W-14666893


jkasturi-sf (approved):
Looks good

